PR #16, opened 2017-08-14
Delete the pull request template

kytrinyx:
I think it is adding unnecessary noise.


kytrinyx:
Did this in a branch/PR as an FYI, but I'm going to go ahead and merge this as soon as it passes CI.

Stargator:
Mmm....maybe :) I wasn't sure since the template issues seemed more related to code PRs

kytrinyx:
> I wasn't sure since the template issues seemed more related to code PRs

Yeah, it's a bit weird right now.

I think that it would be better to see how this goes without it, and then add something back if we want to address specific concerns with contributions.

PR #17, opened 2017-08-14
Exercise Insert Info

Stargator:
Based on the Java template


kytrinyx:
Totally works. Thanks!

PR #21, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

Note that we're aiming for 10 lines and a max-width of 40 columns.
This solution has a max-width of 59, so we may want to adjust it.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,5 @@
+class HelloWorld {
+  void main() {
+    print("Hello World"):
+  }
+}


Stargator:
The example you have looks to be an answer to an exercise. Just want to make sure that's okay.

kytrinyx:
Yeah, that's fine. That said, it would be better to just make this a straight-up hello world without the conditional, I think.

Is this what it would look like?

```
class HelloWorld {
  String hello() => "Hello, World!";
}
```

Stargator:
Normally, there should be a main() function in order for the single dart file to run properly.

So depending on whether you want the main () then it could be:

1)
```dart
class HelloWorld {
  void main() {
    print("Hello World"):
  }
}
```

2)
```dart
class HelloWorld {
  String hello() => print("Hello, World!");

  void main() {
    hello();
  }
}

Based on that this will be based on the first exercise. Then I think what you did will serve that.

kytrinyx:
The goal of the snippet isn't related to the exercises at all, it's really just a way of giving people a quick visual taste of what the language looks like.

I think both of your examples do a grand job of that, however we have a 40 column limit, and example 2 has a width of 43 characters.

So either we name the function `hi()` instead of `hello()`, or we should go with example 1.

I'm in favor of using example 1.

Stargator:
I think example 1 will be fine.

kytrinyx:
@Stargator I've updated the example to use your first suggestion.

Also—I really appreciate your help on this one. Would you by any chance be interested in having commit on this repo, and helping review and merge the occasional pull request? The volume is quite low, and you wouldn't be required to do anything more than you're already doing unless you wanted to.

Stargator:
@kytrinyx I think I would be fine with that then.

kytrinyx:
I appreciate it! Welcome to Exercism :-)

We have some (possibly woefully inadequate) documentation here: https://github.com/exercism/docs/blob/master/maintaining-a-track/README.md If you ever have any questions, please ask them so we can improve the docs!

Stargator:
Now I have something to read this weekend :)

Issue #22, opened 2017-08-28
Tests page has numbering that isn't consistent


Stargator:
The [tests page](http://exercism.io/languages/dart/tests) has numbered bullet points that somehow get reset.

I don't see that when viewing the markdown file in Github, so I am not sure what the process is that interprets the file.

This is also affecting [the Java track's test page](http://exercism.io/languages/java/tests)

Adding an image just for future reference of the original problem:

![redcarpet01](https://user-images.githubusercontent.com/7527155/29997282-646f9e96-8fdd-11e7-8d04-5ba2e92ac448.png)

iHiD:
The markdown parsing code we're using has changed for v2, so whatever turns up here will need to be applied there too.

Stargator:
@iHiD I actually do not see the Tests page for Java in v2. Are you saying there are two different parsing logic for v2 and v1?

Since v2 of Java isn't using the TESTS.md page it's hard to tell if v2 handles v1 the same from where I am.

kytrinyx:
I'm pretty sure that if we fix the markdown for v1 it will also be fixed for v2 (but yes, v2 does markdown parsing differently—better—than v1).

Stargator:
@iHiD @kytrinyx Should this issue be closed and opened in a different repo? Since it seems to concern the markdown parsing, I want to check if this is okay.

Stargator:
[Comment from NobbZ](https://gitter.im/exercism/dev?at=59aadd0bba0f0f6e380d4bcc):

> Also I can confirm, that v2 suffers from the same problem, I posted the dart test-documentation straight into a comment on my own exercises ;)

kytrinyx:
@Stargator Yeah, would you mind posting reproduction steps in https://github.com/exercism/v2-feedback/issues ? (A markdown comment that exhibits the same behavior as the dart test docs would probably be our best bet here)

Stargator:
@kytrinyx I will post the same thing there as well and close this issue.

My findings:

Redcarpet with code formatting as seen in the image in the initial post does not convert to the expected html format. The code formatting breaks the numbering for lists.

I created a jekyll project to test our Markdown library redcarpet. You can find it [a project for redcarpet here](https://github.com/Stargator/markdown-with-redcarpet/tree/red-quotes) and [a jekyll project using kramdown here](https://github.com/Stargator/markdown-with-kramdown).

I will be making the same post in the redcarpet project as well by COB today.

The problem remains when using quotes with redcarpet:
[...]

PR #23, opened 2017-09-02
Fix numbering issue with Markdown as reported

Stargator:
As reported in #22, I am trying to provide a workaround.

Since it seems the code blocks are causing the issue I am going to try some alternatives

I will test these in a jekyll page since this seems to be related to red-carpet, we otherwise won't know how the webpage would turn out.

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -86,7 +86,7 @@ If you get stuck, at any point, don't forget to reach out for [help](http://exer
    ```
    *(Don't worry about the tests failing, at first, this is how you begin each exercise.)*
 
-6. Solve the exercise.  Find and work through the `README.md` guide ([view on GitHub](https://github.com/exercism/dart/blob/master/exercises/hello-world/README.md)).
+5. Solve the exercise.  Find and work through the `README.md` guide ([view on GitHub](https://github.com/exercism/dart/blob/master/exercises/hello-world/README.md)).
 
 Good luck!  Have fun!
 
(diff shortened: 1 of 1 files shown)


Stargator:
No workaround currently found. Though I have noted a few issues.

In the Mac and Linux section, I made a typo and it goes  3,4,6 instead of 3,4,5.

Additionally, I changed the language used for Mac and Linux from batchfile to shell.

Stargator:
@iHiD @kytrinyx 

[According to the Markdown syntax](https://daringfireball.net/projects/markdown/syntax#list), code blocks need to be double spaced. Kramdown is more forgiving in this case. But I have a solution and will create a pull request for java as well as look for any instances of this case in v1.

You can view the working example in [this branch of a example repo](https://github.com/Stargator/markdown-with-redcarpet/tree/red-double-space). Run `bundle install` and `bundle exec jekyll serve`. Assuming you have Ruby, Jekyll, and Bundler installed.

Screenshot as proof. Notice how Windows, which in this case is double indented, renders properly and macOS, which remain untouched, does not:

windows-vs-mac

Stargator:
Issue stems from: https://github.com/exercism/v2-feedback/issues/105

kytrinyx:
Thanks for digging into this.

The one thing I always do with markdown is use all the same number for numbered lists (usually 0 or 1). That way it should do the numbering by itself, and it's easy to insert new list items between existing items or reorder the items without having to edit the number itself.

If the indentation is correct, markdown _should_ handle the numbering automatically.

Stargator:
@kytrinyx I'll make that change. I'll make all the numbers "1".

Issue #27, opened 2017-09-04
Rework the Hello World exercise test suite to be a template for other exercises


Stargator:
Stemming out of #24, based on the discussion there the plan is to rework the Hello World exercise's test file to use typing since Dart 1.0 is optional typing so our test files are prepared for Dart 2.0.

Additionally, some test suites are structured differently. For now, I think the first 5 exercises should not have any groups. Then we can slowly add *and* explain uses of the group function in the second set of exercises.

Once this is done for Hello World, other issues will be created to update the other exercises and follow suite.

Stargator:
Pull Request was opened for this, so the conversation can be continued there.

PR #29, opened 2017-09-04
Revise Hello World to serve as a simple example for future test suites

Stargator:
Addresses #27 

I also bumped the test version to 0.12.24+1. Not sure if we would be okay with setting a [version constraint](https://www.dartlang.org/tools/pub/dependencies#version-constraints) to the test framework can go up to, but not including, 0.13.0.

We can do this by setting the version for test to be `<0.13.0`, then that gives us some breathing room and we don't have to think about updating the dependencies so often.

### DIFF
--- a/exercises/bob/pubspec.lock
+++ b/exercises/bob/pubspec.lock
@@ -42,13 +42,13 @@ packages:
       name: cli_util
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.1.1"
+    version: "0.1.2"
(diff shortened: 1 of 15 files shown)


kytrinyx:
> then that gives us some breathing room and we don't have to think about updating the dependencies so often.

I think that sounds wise.

Stargator:
@kytrinyx Since you approve bumping the dependency, let me go ahead and do that for the other exercises in this branch.

Issue #34, opened 2017-09-04
Create a script to be used by Travis for testing exercises against reference solution


Stargator:
Stemming out of the conversation at #3.

We need a script that Travis can execute, along with the current scripts/commands, that will do the following:
* Copy the exercises directory to a temp directory.
* Go into each exercise and modify the test suite's import path.
   * The modification will be replacing the normal import for "bob.dart", for example, to "example.dart". The example.dart file in each exercise acts as our reference solution. The answer key to the problems in the exercise.
* Run `pub run test` on each exercise.

Stargator:
Additionally, you can look at the [Track config file](https://github.com/exercism/docs/blob/master/language-tracks/configuration/track.md) to see what the reference solution and test file patterns are.

The Track's config file can be [found here](https://github.com/exercism/dart/blob/master/config.json)

kytrinyx:
It would also need to change the test suite so that none of the tests are skipped.

Stargator:
@kytrinyx It seems the mighty @SuperPaintman has already thought about this. The test file in the root test directory already looks at the command and replaces the use of the normal file with the reference solution and then runs all the tests (no skipping).

You can see this by running `pub run test` from the root directory of our project. You'll see all the tests are run and they are successful. I just need to add that step to Travis and we'll be set.

I'm also going to check that the right exercise is properly labeled before showing the test results.

kytrinyx:
> It seems the mighty @SuperPaintman has already thought about this

This is great news! Alright, it looks like we're in great shape for CI, then.

Stargator:
Closing since this was implemented.

PR #35, opened 2017-09-05
Run all tests in Travis and revise dependency versions

Stargator:
Implementation of #34. Add a command to travis to run the tests.

@SuperPaintman already had a script to run all tests. But since that was also a test, the final results were not separated from the test results of the last exercise. So I added a statement in between.

Additionally updated the repo's package dependencies to reflect the changes to versioning we made for the exercises.

***Note to Reviewer***: No changes to Travis Settings requested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,18 +9,16 @@ dart:
 dart_task:
   - test: --platform vm
 
+branches:
+  only:
+  - master
+
 cache:
   directories:
     - $HOME/.pub-cache
 
 before_script:
-  - |
-    unformatted="$(pub run dart_style:format -l 120 -n .)"
-    if [ ! -z "$unformatted" ]; then
-      echo "Files are unformatted:"
-      echo "$unformatted"
-      exit 1
-    fi
+  - bin/check_formatting
 
 script:
   - bin/fetch-configlet
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,4 @@ before_script:
 script:
   - bin/fetch-configlet
   - bin/configlet lint .
+  - pub run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,6 @@ dart:
 dart_task:
   - test: --platform vm
 
-branches:
-  only:
-  - master
-
 cache:
   directories:
     - $HOME/.pub-cache
--- a/test/exercises_test.dart
+++ b/test/exercises_test.dart
@@ -64,10 +64,17 @@ Running tests for: $packageName
   await runCmd(["cp", "test/${packageName}_test.dart", "test/${packageName}_test.dart.bu"]);
   try {
     for (dynamic cmds in [
-      ["cp", "lib/example.dart", "lib/${packageName}.dart"], // Replace main file with example
-      ["sed", "-i", "-e", "s/\\bskip:\\s*true\\b/skip: false/g", "test/${packageName}_test.dart"], // Enable all tests
-      ["pub", "get"], // Pull dependencies
-      ["pub", "run", "test"] // Run tests
+      /// Replace main file with example
+      ["cp", "lib/example.dart", "lib/${packageName}.dart"],
+
+      /// Enable all tests
+      ["sed", "-i", "-e", "s/\\bskip:\\s*true\\b/skip: false/g", "test/${packageName}_test.dart"],
+
+      /// Pull dependencies
+      ["pub", "get"],
+
+      /// Run tests
+      ["pub", "run", "test"]
     ]) {
       await runCmd(cmds);
     }
@@ -96,7 +103,6 @@ Future runAllTests() async {
 Running tests for: $packageName
 ================================================================================
 """);
-
 }
 
 void main() {
--- a/test/exercises_test.dart
+++ b/test/exercises_test.dart
@@ -27,14 +27,32 @@ Future runCmd(List<String> cmds) async {
   assert(res.exitCode == 0);
 }
 
+Future getPackageName() async {
+  final pubspec = new File("pubspec.yaml");
+
+  final packageName = loadYaml(await pubspec.readAsString())["name"];
+
+  return packageName;
+}
+
+Future locateExercismDirAndExecuteTests() async {
+  final exercisesRootDir = new Directory("exercises");
+
+  assert(await exercisesRootDir.exists());
+
+  final exercisesDirs = exercisesRootDir.listSync().where((d) => d is Directory);
+
+  for (dynamic dir in exercisesDirs) {
+    await runTest(dir.path);
+  }
+}
+
 Future runTest(String path) async {
   final current = Directory.current;
 
   Directory.current = path;
 
-  final pubspec = new File("pubspec.yaml");
-
-  final packageName = loadYaml(await pubspec.readAsString())["name"];
+  dynamic packageName = await getPackageName();
 
   print("""
 ================================================================================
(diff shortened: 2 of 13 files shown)


kytrinyx:
> Before this PR may be merged, I recommend changing the settings on travis-ci.org so that builds occur for branch updates.

Would you say more about this? I've explicitly disabled building for branches, because when it is turned on it takes twice the resources. We only have 5 concurrent builds, and with 94 repositories (many of which are active) it can take a long time before builds run.

Stargator:
@kytrinyx I recommend it because I don't think a build is done when someone commits directly to the master branch.

Do note the Travis config only builds for the master branch.

Stargator:
@iHiD  @SuperPaintman would either of you be able to review this?

iHiD:
@stargator I'm not the right person for this - sorry. 

Best to wait for @kytrinyx to be back or @SuperPaintman :)

Stargator:
Ok, thanks. There's no one else?

iHiD:
I don't know - sorry. My domain is product and website. Everything else I'm less sure of. From looking at the contributors to this repo, it doesn't look like there's anyone else particularly knowledgable on dart. @nobbz is generally helpful and might know who would be good to help?

NobbZ:
> I recommend [activating travis on push] because I don't think a build is done when someone commits directly to the master branch.

There shouldn't be any pushes directly to master. Only PRs against it. But to be honest, I do often forget to switch branches as well ;)

> Do note the Travis config only builds for the master branch.

As I read the `.travis.yml` there should be a build triggered on every branch.

Disclaimer: I'm not sure how exactly travis is set up besides the things I can read from the yml. Also I'm not able to even look at the configs at the travis web page nor in the github repository. AFAIK those are only accessible by @kytrinyx 

Besides of that, I do not know anything about dart and therefore can't review the PR itself nor could I merge it. Only @exercism/dart are able to (+ @kytrinyx)

Stargator:
Thanks @iHiD and @NobbZ.

@NobbZ I'm on the exercism/dart team. But don't want to merge things on my own without some kind of review.

Regarding .travis.yml I was referring to the change I made in this PR:
```yaml
branches:
  only:
  - master
```
I made the change so that we don't worry about builds for each and every branch. Just the made one.

I understand the logic that if all the work is added via PRs then the travis builds for it will track breaking changes. But we can't control everyone that may get on the team and so we should have builds for changes to master to ensure that the health of that branch is tracked.

NobbZ:
As far as I remember, @kytrinyx wanted to "protect" master against direct pushes anyway.

Also most contributors will stay outside of the maintainers team and therefore be only able to commit via PRs. And people that are in the tracks maintainer team, of course need to be careful to not push directly to master (if its not protected already).

In general I'm not sure how the branches-setting in travis-file does behave on PRs. Eg will it test PRs against master? What if we have some longrunning branch preparing for the next big overhaul and we do PRs against that? In my opinion, we need to lock down direct commits to the repositories branches as best as possible while doing anything via PRs. This is to enforce review process and proper testing.

I try to avoid pushing to upstream as much as possible when working in the erlang track, but to do anything on my fork from which I PR then. Even though I often just merge instantly because I'm the sole maintainer left :(

Stargator:
That's true, @kytrinyx does want to use PRs in order to encourage feedback. I don't have access to the settings of the repo, so I can't check the settings. But I **feel** like I have pushed to master at least once when I was new to the team and just doing hotfixes.

Anyway, there's some other things I can get started in the mean time.

kytrinyx:
> I don't think a build is done when someone commits directly to the master branch.

Agreed. I turned off the "build branch on push" option in Travis explicitly. Exercism has 90+ repositories, many of which are active. Travis allows 5 concurrent builds. If we turn on the build branch option, then we double the number of builds that we do.

> kytrinyx does want to use PRs in order to encourage feedback. 

Agreed, I don't want us pushing directly to master.

This is not just about feedback, it's about visibility. Pushing to master hides what's going on from other contributors and maintainers. If you have a trivial change, then please make a branch, create a PR, and merge it yourself.

.travis.yml
```diff
@@ -9,19 +9,18 @@ dart:
 dart_task:
   - test: --platform vm
 
+branches:
+  only:
+  - master
+
```
kytrinyx: This is unlikely to build, and frankly I'd rather not have pushes to master at all. I'm going to try to wrap up the discussion about protecting master via repo settings and automate this change across all the tracks.

Stargator:
@kytrinyx will remove the branch setting

Stargator:
@kytrinyx .travis.yml updated with removal of the branch settings

Stargator:
@fwip or @kabiir Would either of you feel up to reviewing this? Let me know if you have any questions.

kabiir:
@Stargator Travis is still new to me, nonetheless I'll look into it and see if I can provide any helpful feedback as per the dart part a quick look indicated everything is okay but I'll look deeper after I explore the Travis part.

test/exercises_test.dart
```diff
@@ -45,11 +68,18 @@ Running tests for: $packageName
   await runCmd(["cp", "lib/${packageName}.dart", "lib/${packageName}.dart.bu"]);
   await runCmd(["cp", "test/${packageName}_test.dart", "test/${packageName}_test.dart.bu"]);
   try {
-    for (var cmds in [
-      ["cp", "lib/example.dart", "lib/${packageName}.dart"], // Replace main file with example
-      ["sed", "-i", "-e", "s/\\bskip:\\s*true\\b/skip: false/g", "test/${packageName}_test.dart"], // Enable all tests
-      ["pub", "get"], // Pull dependencies
-      ["pub", "run", "test"] // Run tests
+    for (dynamic cmds in [
```
kabiir: The `cmds` variable should be `List<String>` . 
Use `dynamic` as a function return type when the function returns a `string` or a `map` or `whatever data type`.
You can also call any function on a `dynamic` variable assuming it exists. 
```dart
(foo as dynamic).whatever(); //valid. compiler won't check if whatever() exists
(foo as var).whatever(); //illegal. var is not a type
```
Use `var` when for a variable you expect its data type to change during its lifetime.
e.g.

```dart
dynamic myFunc(int a){
  if (a == 1) return "String";
  if (a == 2) return [1,2,3]
  return null
}

/// Here it becomes a String.
var myVariableDataTypeVariable = myFunc(1);
/// Here it becomes a List.
var myVariableDataTypeVariable = myFunc(1);
/// Here it becomes null.
var myVariableDataTypeVariable = myFunc(1);

```

In conclusion previously `var` was fine. But we should explicitly make it of type `List<String>`.

test/exercises_test.dart
```diff
@@ -27,14 +27,37 @@ Future runCmd(List<String> cmds) async {
...
+Future locateExercismDirAndExecuteTests() async {
+  final exercisesRootDir = new Directory("exercises");
+
+  assert(await exercisesRootDir.exists());
+
+  final exercisesDirs = exercisesRootDir.listSync().where((d) => d is Directory);
+
+  /// Sort directories alphabetically
+  final sortedExerciseDirs = exercisesDirs.toList();
+  sortedExerciseDirs.sort((a, b) => a.path.compareTo(b.path));
```
kabiir: Why is there a need to sort the list?

test/exercises_test.dart
```diff
@@ -27,14 +27,37 @@ Future runCmd(List<String> cmds) async {
...
+
+  for (dynamic dir in sortedExerciseDirs) {
```
kabiir: Same point for using `dynamic`. See below comment
Use `FileSystemEntity`, since we are actually calling a getter `path` in the loop it's important that it actually exists.
The compiler won't check if it's a `dynamic` variable.

test/exercises_test.dart
Stargator: No need, just thought it would be easier to look for tests in the output on travis-ci  or when running locally.

Stargator:
@kabiir Made changes as requested. There was also use of dynamic for packageName, so I changed that too.

kytrinyx:
I'm so sorry I lost track of this one, @Stargator. Thanks so much for taking the time to work through this! 🌷

Stargator:
No worries. I got support :)

Stargator:
No worries. I got support :smile:

PR #44, opened 2017-10-02
Some Test suites accidentally reference the example.dart file

Stargator:
Two test suites are referencing "example.dart" which is the solution to the given problem. Nothing is leaked to the user, because the file is not included when users fetch the problem.

But we shouldn't include errors into our user's test suites, so we should remove it.

@exercism/dart or @fwip care to check me over?


fwip:
Looks good. 👍

PR #46, opened 2017-10-03
Add Word Count exercise

kabiir:
Shows usage of regular expressions and forEach method.
Passes all specified tests.
I couldn't find how to get the uuid for this exercise to enter it in the config.json file.


kabiir:
Can anyone tell me how would i format my code so that the travis ci build passes

Stargator:
@kabiir `pub run dart_style:format -l 120 -w` Just be sure to include the path to the exercise after the `-w`

Thanks for the work!

exercises/word-count/lib/example.dart
```diff
@@ -0,0 +1,25 @@
...
+    // \\w+'\\w  => matches any number of letters then "'" and then a single letter
+    //  e.g. don't
+    // \\d+ matches one or more digits
+    List<String> allwords = new List();
+    var wordMatches =
+        new RegExp("\\w+'\\w|\\w+|d+").allMatches(input.toLowerCase());
+    for (var match in wordMatches) {
+      allwords.add(match.group(0));
+    }
+    // print(allwords);
```
Stargator: The print statement can be removed

exercises/word-count/lib/example.dart
```diff
@@ -0,0 +1,25 @@
...
+    // Lists aka "arrays" have a forEach method that applies a function to
+    // each element of the list.
+    // initialize the Map with all words and there counter to 0
+    allwords.forEach((singleWord) => count[singleWord] = 0);
+    // count all words and set there counter accordingly
+    allwords.forEach((singleWord) => count[singleWord] = count[singleWord] + 1);
+    return count;
+  }
+}
```
Stargator: Can you convert all the double slash comments to ///? If we ever do documentation, that's the Dart preference I think.

exercises/word-count/lib/word_count.dart
```diff
@@ -0,0 +1,3 @@
+class WordCount {
+  // Put your code here
```
Stargator: Though, this is fine.

exercises/word-count/pubspec.yaml
```diff
@@ -0,0 +1,3 @@
+name: 'word_count'
+dev_dependencies:
+  test: "^0.12.24+8"
```
Stargator: Can you change the test version to <0.13.0 to be consistent with the other exercises?

exercises/word-count/test/word_count_test.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:test/test.dart';
+import 'package:word_count/word_count.dart';
+
+void main() {
+  final mywordcounter = new WordCount();
+  group("WordCount", () {
+    test("count one word", () {
```
Stargator: Add a space between mywordcounter and group.

In order to following the Dart Style Guide, rename mywordcounter to [camelCase syntax](https://www.dartlang.org/guides/language/effective-dart/style#do-name-other-identifiers-using-lowercamelcase).

exercises/word-count/test/word_count_test.dart
```diff
@@ -0,0 +1,56 @@
...
+            "joe": 1,
+            "can't": 1,
+            "tell": 1,
+            "between": 1,
+            "large": 2,
+            "and": 1
+          }));
+    });
+  });
+}
```
Stargator: I would request the tests to be separated into additional groups (see Anagram). Where it makes logical sense to you.

kabiir:
@Stargator, I'll be finishing this PR in couple of hours. My time zone is different so I'll start working on other exercises a bit late than you would expect.

kabiir:
@Stargator Please review again, i've refactored the code. i was hoping to add default parameter to the word_count.dart but didn't know if i should. Beginners won't know where to start from so i think i should add 

`
void countWords(String inputString){
  // Put your code here
}
`

or
additionally 

`
Map<String, int> countWords(String inputString){
  // Put your code here
}
`

The later might force them to go for a specific approach.
So which one should i add?

Stargator:
Leave the class empty. It's up to the user to determine a solution. They already know a function named `countWords` is expected. It wouldn't add anything.

I recommend setting the value of `myWordCounter.countWords` to a final List<Map, String>.

Sort the topics alphabetically.

In the config.json, I recommend moving the word-count exercise after bob. Update the "unlocked_by" parameter respectively for both.

exercises/word-count/test/word_count_test.dart
```diff
@@ -0,0 +1,58 @@
...
+  group("WordCount: Simple Tests - ", simpleTests);
+  group("WordCount: Ignore special characters - ", ignoreSpecialCharacters);
+  group("WordCount: Works with numbers too - ", notJustWords);
+  group("WordCount: Edge case - ", edgeCases);
+}
+
+void simpleTests() {
+  test("count one word", () {
+    expect(myWordCounter.countWords("word"), equals({"word": 1}));
+  });
```
Stargator: Whatever is the first test as defined in the canonical.json, should have have `}, skip: false);` at the end, all other tests should have `}, skip: true);`

Stargator:
Looking good, @kabiir! Sorry for the back a forth, there's a lot to consider and review.

kabiir:
@Stargator it's fine, I'm enjoying dart.

kabiir:
> I recommend setting the value of myWordCounter.countWords to a final List<Map, String>.

I don't understand what you mean by that.

Stargator:
@kabiir Sorry, I can see how I didn't word that correctly.

I meant to say, I recommend having the value returned by `myWordCounter.countWords` set to a variable of `final List<Map, String>`. The intent is to take some of the logic out of the expect so that users can read just that one line. With the result being the test suite is a bit more readable.

kabiir:
@Stargator oh! So you mean to save the value to a variable first and then compare it with the expected value. So the variable will be of type `Map<String, int>`. That would mean I would need to repeat the same in each test case & hence repetition of code.
Also if I do that, the user will get too much hint, I think they should explore and learn on there own about Maps. They might learn about Maps and implement it too but the tests would still fail and they'll slowly realize that the function return type also needs to be the same map. After all tests are supposed to be that way. If nothing else they can pull the example. This exercise has topics maps and lists. This is what they are supposed to learn on there own. Your thought?

Stargator:
@kabiir I can see your point, but I think we can be easier on the core exercises.

But I'll leave it to you.

Issue #40, opened 2017-10-01
Hamming exercise needs to be updated due to changes in specification


Stargator:
The specification for the Hamming exercise can be [found here](https://github.com/exercism/problem-specifications/tree/master/exercises/hamming).

There are a few exercises that were either added or changed.

The changes should occur in these files:
* exercises/hamming/test/hamming_test.dart
  * This file will reflect the content of canonical-data.json
* exercises/hamming/lib/example.dart
  * This file will have to be updated in case the changes in hamming_test.dart render the solution outdated with the new/changed tests.

To run the tests, execute `pub run test` from the root directory of exercism/dart

jerold:
Hi @Stargator would you be able to clarify what has changed about the Hamming spec?

Stargator:
@jerold No problem.

[In the canonical json](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json), some tests were not implemented. Feel free to keep the current tests and just merge in these other tests.

And ignore any that may seem duplicates of the current batch of tests.

The first case (empty strands) from the json is not implemented.

And the following cases:
* non-unique character in first strand
* non-unique character in second strand
* same nucleotides in different positions
* disallow first strand longer
* disallow second strand longer

jerold:
Thanks @Stargator I've pushed the revert and test changes.

Stargator:
Related Pull Request #48 was merged

PR #48, opened 2017-10-04
Revise Hamming tests with missing cases from specification

jerold:
## Issues
https://github.com/exercism/dart/issues/40

Updating tests for Dart hamming distance example.


exercises/hamming/lib/hamming.dart
```diff
@@ -1,3 +1,5 @@
 class Hamming {
-  // Put your code here
+  int compute(String a, String b) {
+    // Put your code here
+  }
```
Stargator: This can be reverted. The users will know it needs to be implemented from the tests. Best let them decide the details like the types used.

Stargator:
@jerold Just checking in, how do you feel about my feedback?

jerold:
@Stargator Sorry about the delay. I was out of town last weekend.  I'm going to clean this up today.

Stargator:
@jerold No worries. Thanks for remembering us 😸

jerold:
@Stargator restored the previous set of tests and implemented the 6 you've asked for. Run the new tests against the example hamming implementation and all pass. Ready for re-review.

jerold:
@Stargator Sorry for the churn. The previous commit by itself may look a little confusing. I effectively undid the previous commit (where I'd reordered/renamed the tests to exactly match the canonical-data.json file). Looking at the diff for the whole pr currently should show 6 tests added:
- [ ] empty strands
- [ ] non-unique character in first strand
- [ ] non-unique character in second strand
- [ ] same nucleotides in different positions
- [ ] disallow first strand longer
- [ ] disallow second strand longer

Do you mean to say the bottom 5 of those should be skipped, or should all tests in the file, minus the top test be skipped? Are these tests brought online one by one somehow as the coder works on their implementation? I'm curious what the process is for these tests that required them to be skipped initially? Are these part of multi-part tutorial?

Stargator:
@jerold The tests should be done one at a time. Generally, the top test is the first test (we don't skip the first test). Then all the tests that follow it will be appended with `}, skip: true);`

> should all tests in the file, minus the top test be skipped?

Yes

The process is that we are signalling to the user that the first problem for them is the only test that is failing, the first test. Then they solve that problem by passing the first test.

Then they stop skipping the second test, see that it fails and work to resolve it. Then repeat the cycle until all the tests pass.

exercises/hamming/test/hamming_test.dart
```diff
@@ -5,11 +5,43 @@ void main() {
...
+      final res = hamming.compute("AGA", "AGG");
+
+      expect(res, equals(1));
+    }, skip: true);
+
+    test("non-unique character in second strand", () {
+      final res = hamming.compute("AGG", "AGA");
+
+      expect(res, equals(1));
+    }, skip: true);
```
Stargator: I can't see a reasonable difference between the two tests.

For "non-unique character in first strand", if the first strand was "AAG" and the second strand was "AAA", then I would understand the problem.

Then reverse it for "non-unique character in second strand".

jerold:
@Stargator give this a look. I've translated the full test suite described in [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json) to dart tests. Each test has a counterpart. Same descriptions, cases, expectations. The diff is a little ugly, but take a look at the current raw test file next to canonical, and I think you'll like it. Otherwise I can revert, but in that case I would think a change really should be made to canonical first, and the tests then made to match here in hamming_test.dart.

Stargator:
@jerold Understood. But do understand, we don't have to follow the canonical-data 1 to 1.

So I opened a PR in the problem-specifications https://github.com/exercism/problem-specifications/pull/953, feel free to comment.

PR #50, opened 2017-10-04
Add Space Age exercise

kabiir:
I skipped adding space-age to config.json on purpose, because I'm not sure where it should be added, what difficulty it should have, topics and unlocked by.

This exercise introduces usage of setters and getters and shows how to pass "data" to an instance of a class and how to retrieve it without using any function. It also introduces enumerations (`enum`) data type and `switch` conditional flow.

The [TOPICS.txt](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) doesn't seem to have any entries for getters & setters


exercises/space-age/test/space_age_test.dart
```diff
@@ -0,0 +1,50 @@
...
+    myAge.inSeconds = 3210123456;
+    myAge.planet = Planets.Uranus;
+    expect(myAge.age, equals(1.21));
+  }, skip: true);
+  test("Neptune", () {
+    myAge.inSeconds = 8210123456;
+    myAge.planet = Planets.Neptune;
+    expect(myAge.age, equals(1.58));
+  }, skip: true);
+}
```
Stargator: Could you space out the tests? It looks a bit clumped together.

Stargator:
It definitely introduces getters/setters. But not so convinced about enums, since it is up to the user how they implement Planet it could easily be a class with fields for each planet of a certain value.

And likewise about switch, it's there and people can use it, but the tests don't encourage it one way or another. Users could wind up doing nested ifs.

And that's **all okay** :smile:

About the topics, I think it touches on the following, which you can widdle down as you see fit.

* conditionals
* logic
* mathematics

Additionally, if you want getters/setters to be topics, open a PR for the exercism/problem-specifications :)

kabiir:
@Stargator People who'll start learning dart will most probably use the official dart editor or perhaps vscode. These editors can suggest the values for switch cases. They can also automatically create cases for `switch ` if the parameter for switch statement is an enum. The point behind using `enum` is that, it's values are readily available you don't need to instantiate it or call fancy functions or maps or lists. Enumeration are best suited for a finite set of values which have case sensitive importance too. The only other option is to store planet names as strings in either maps, lists or a fancy class. But all these need to be immutable too so that the end user doesn't accidently change its value. And the input string for the function will also need to match the case. This all can be avoided with enum.

kabiir:
```dart
class Planets{
any-data-type get Mercury => "something" ;
}
```

This is another accepted way that the test cases will allow to pass . i.e. Having a class named Planets with an only getters for all planets.

However I can reduce the program to mere few lines of code too by creating constants for all the planets and setting there value as given in the exercise Readme file. This will also allow the user to not actually use switch cases or `if` statements at all.

Instead of the whole switch case it can be reduced to just one line of code. But that's just another way of writing this program. My goal was to introduce as many reasonable concepts as possible.

Stargator:
> @Stargator People who'll start learning dart will most probably use the official dart editor or perhaps vscode. These editors can suggest the values for switch cases.

We can't know what they are using, so we cannot anticipate they will get suggestions.

Also, I see that you need to update the branch to have the same commits as `master`.

kabiir:
@Stargator I still need opinions about this exercise (config.json) I think phone numbers is next. And this should be after phone numbers.

Stargator:
For the config.json:

Core: yes

Suggested Topics:
"mathematics"
"floating_point_numbers"

Let me know if other topics were on your mind.

Difficulty: 1

unlocked by word count
 * Also, update bob to be unlocked by this.

Stargator:
This PR in the problem specifications as approved: https://github.com/exercism/problem-specifications/pull/946

Recommend when it gets merged, if it impacts us,that this ticket is updated.

Also, it doesn't seem like there was a response to [the previous suggestions](https://github.com/exercism/dart/pull/50#issuecomment-335013453).

exercises/space-age/lib/example.dart
```diff
@@ -0,0 +1,69 @@
...
+        break;
+      case Planets.Venus:
+        this._ageInYears = (ageInSeconds / earthYearInSeconds) / 0.61519726;
+        break;
+    }
+    ;
+  }
+
+  /// Returns `double` rounded of to [fractionDigits].
+  double roundOf(double number, {int fractionDigits = 2}) => num.parse(number.toStringAsFixed(fractionDigits));
```
kabiir: Forgot to actually use this method!

kabiir:
@Stargator I think I should add a tolerance of `0.5` to all tests. So that the students don't have to round of their answer

Stargator:
The [specification of this exercise was updated](https://github.com/exercism/problem-specifications/pull/1104) to follow a new definition of the schema. You may want to double check there weren't any additional changes.

devkabiir:
https://github.com/exercism/problem-specifications/pull/1104 Is only updating schema, I don't think this needs updating.

exercises/space-age/lib/example.dart
Stargator: Are you going to drop this method then? Or are you going to use it?

exercises/space-age/pubspec.lock
```diff
@@ -0,0 +1,293 @@
+# Generated by pub
```
Stargator: This file can be removed as we recently started dropping them in favor of having the users generate lock files with more up to date dependencies.

exercises/space-age/test/space_age_test.dart
```diff
@@ -0,0 +1,57 @@
...
+
+final SpaceAge myAge = new SpaceAge();
+
+void main() {
+  group("SpaceAge: Planet - ", ageOnPlanetTests);
+}
+
+ageOnPlanetTests() {
+  test("Earth", () {
+    myAge.inSeconds = 1000000000;
```
Stargator: The canonical-data.json calls for the values for planet and seconds to be passed to a method called age.

Additionally, the value for planet would be of type String. The enums can still be used in the SpaceAge class, but to remain consistent with the canonical-data.json, the test gives the planet name as a String.

Stargator: Thinking about it now, I think it's fine to leave the enums in the test file.

Stargator:
@devkabiir Once the above is done, this PR can be approved.

devkabiir:
@Stargator Hi!, it's been a while, things seem to be going well here, I'll update my PRs starting with this one. I'm going through all of my emails from this repo and it seems a lot has changed. I'm hoping to close this one in a day or two.

devkabiir:
@Stargator ready for another review

config.json
```diff
@@ -80,6 +80,20 @@
...
+      "slug": "space-age",
+      "uuid": "c8bf7da7-d26d-48e8-a83b-1296972d2376",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "enumerations",
+        "floating_point_numbers",
+        "integers",
+        "mathematics",
```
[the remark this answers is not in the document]
kytrinyx: @Stargator Would you mind using backticks (e.g. `@kytrinyx`) if you're mentioning me but don't actually need my attention?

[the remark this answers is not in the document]
Stargator: I don't mind. Sorry.

devkabiir:
Yes, it is @jvarness. I'm planning rewrite the history.

exercises/space-age/lib/example.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:meta/meta.dart';
```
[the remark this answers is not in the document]
devkabiir: This reminds me, there should be an intro to dart annotations somewhere.

devkabiir:
@Stargator @jvarness 
Update: 
changed the [example.dart](https://github.com/exercism/dart/pull/50/files#diff-31109633f2223dae26ac844b058f3f24) implementation
updated test to comply with [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/space-age/canonical-data.json#L7) descriptions and input types
added [`analysis_options.yaml`](https://github.com/exercism/dart/pull/50/files#diff-7eaa5baf7a385dcd2fe1244b7ad12dd6)
updated [`pubspec.yaml`](https://github.com/exercism/dart/pull/50/files#diff-2af5cb0112dc054d4fe51029c87a4cd8)
changed [`config.json`](https://github.com/exercism/dart/pull/50/files#diff-06b2d3b23dce96e1619d2b53d6c947ec) topics and difficulty

exercises/space-age/pubspec.yaml
```diff
@@ -0,0 +1,5 @@
+name: 'space_age'
```
Stargator: This has CRLF line-endings. Could you change it to LF?

It's better for git to store files with LF and then on the client-side you can set your config to convert it for your environment to CRLF if needed.

devkabiir: Interesting, I have LF ending set, even in windows, for all of my editors. This has never been a problem. I copied pubspec from another exercise, don't remember which one, that's probably the culprit. I'll check if any other file has that.

Stargator: Do you see it too? If it's just on my machine, then we can ignore it.

Stargator:
@devkabiir You are good to rebase and then merge this in!

exercises/space-age/pubspec.yaml
devkabiir: Actually, that red marker means there is no extra line at the end. And I think it's a standard to have an extra empty line ending the file. I may have used a code formatter which removed the line. It doesn't seem to be an actual problem for git. @Stargator

devkabiir: Learned something new today, it seems most Unix tools expect a file to end with a new line so that when they display its contents in the console they don't have to worry about differentiating the file contents with actual log/status messages. By having a new line the tools don't add any extra line and the output is distinguishable automatically.

Stargator: @devkabiir I added a commit that changes the line ending. I wasn't referring to a red marker. Feel free to fixup my commit into yours, so it's one commit.

I took a screenshot of the change to highlight it, the left side is CRLF (Windows) and the right is LF (Mac, Linux):

screen shot 2018-08-24 at 8 35 11 am

PR #54, opened 2017-10-06
Add Binary Search Tree exercise

devkabiir:
This exercise:
- Focuses on implementation of a data structure using classes.
- explains how privacy (private properties and methods) works in dart.
- has resources and tips section in readme.md.
- has simple tests.

I left the config.json untouched on purpose.

### DIFF
--- a/exercises/binary-search-tree/README.md
+++ b/exercises/binary-search-tree/README.md
@@ -1,0 +1,62 @@
+# Binary Search Tree
+
+Insert and search for numbers in a binary tree.
+
+When we need to represent sorted data, an array does not make a good
+data structure.
+
+Say we have the array `[1, 3, 4, 5]`, and we add 2 to it so it becomes
+`[1, 3, 4, 5, 2]` now we must sort the entire array again! We can
+improve on this by realizing that we only need to make space for the new
+item `[1, nil, 3, 4, 5]`, and then adding the item in the space we
+added. But this still requires us to shift many elements down by one.
+
+Binary Search Trees, however, can operate on sorted data much more
+efficiently.
+
+A binary search tree consists of a series of connected nodes. Each node
+contains a piece of data (e.g. the number 3), a variable named `left`,
+and a variable named `right`. The `left` and `right` variables point at
+`nil`, or other nodes. Since these other nodes in turn have other nodes
+beneath them, we say that the left and right variables are pointing at
+subtrees. All data in the left subtree is less than or equal to the
+current node's data, and all data in the right subtree is greater than
+the current node's data.
+
+For example, if we had a node containing the data 4, and we added the
+data 2, our tree would look like this:
+
+      4
+     /
+    2
+
+If we then added 6, it would look like this:
+
+      4
+     / \
+    2   6
+
+If we then added 3, it would look like this
+
+       4
+     /   \
+    2     6
+     \
+      3
+
+And if we then added 1, 5, and 7, it would look like this
+
+          4
+        /   \
+       /     \
+      2       6
+     / \     / \
+    1   3   5   7
+
+## Resources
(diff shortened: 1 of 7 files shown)


exercises/binary-search-tree/lib/binary_search_tree.dart
```diff
@@ -0,0 +1,3 @@
+class BST {
+  // Put your code here
+}
```
Stargator: Recommend renaming it `BinarySearchTree`

exercises/binary-search-tree/README.md
```diff
@@ -0,0 +1,107 @@
...
+  A aClass = new A();
+
+  testB() {
+    _private = 'Hello';
+    print('String value: $_private'); // Hello
+    print('I cannot access private');
+    print(aClass._private); // Error
+  }
+}
+```
```
Stargator: The example could be shorten to just two classes and no main(). By this point, the users should know how to utilize classes, so if they want to run the code they can.

Stargator:
Given comments in https://github.com/exercism/problem-specifications/issues/555#issuecomment-335044714

I think we should figure out how to order this. It's a complicated problem compared to the rest. I think we will need more exercises that can build up to this.

I'm fine with adding it, but we be mindful of the varying difficulties and balance them out as much as we can.

kabiir:
@Stargator I've made a PR for adding the required canonical data but it's still open and needs opinions. I'd keep this PR open until that one is resolved. And then reorder the tests accordingly. Apart from tests I'd like  your review on the implementation.

It's been a hectic week, it's Diwali season in India. Diwali is in 4 days there's all sorts of preparations pending. So my contributions will be slow this week.

Stargator:
No worries. Thanks.

jvarness:
@Stargator @devkabiir does it make more sense to do other exercises first that involved data structures but are less complicated than binary trees?

Are there available exercises to implement that deal with like... Maps or Lists?

Stargator:
@jvarness Probably, we may even already have some that could use List in the solution. But we haven't looked at specific ones that deal with those types explicitly in the problem.

Also, most of the proposed exercise implementations are not from some high-level plan, it's just whatever each person decided to spend time on. We could hold off on the binary tree implementations until we add a few more exercises that may make t easier for newcomers. We just have't had that discussion, yet. This PR could serve well in that regard.

If interested, [this repo contains all the specifications for current exercises](https://github.com/exercism/problem-specifications) that can be applied to any track. Each exercise should have a defined specification before a track implements it.

devkabiir:
This probably needs a full re-write (tests and example) as per updated canonical-json and all other major changes, I'll work on this probably next weekend.

devkabiir:
Rewrote it compliant with the new canonical-data.json. I left config.json untouched on purpose. We should discuss the that.
@Stargator , @jvarness

devkabiir:
Made a few notable changes:
1. Usage of null-aware operators
2. Updated code docs
3. Stricter types on `BinarySearchTree` and `Node` classes (removed `dynamic`)
4. Moved `assert` statements to constructor bodies. Keeping it in initializer seems to not work in dart 1.24.3

Stargator:
@devkabiir From the root directory of the repo, I ran `pub run dart_style:format -l 120 -n exercises/binary-search-tree/`

And this is the response I got:
`test/binary_search_tree_test.dart`

I was using Dart 2.0, but 1.24.3 should work too. Though, Dart 1.24.3 didn't give me a response, must be due to differences in the SDK.

Stargator:
Hey @devkabiir do you think you can complete this by April 30th? That's the deadline for the New Batch of Exercises milestone

devkabiir:
Closed in #187

Issue #31, opened 2017-09-04
For the Gigasecond exercise, introduce the concept of "setup" and "teardown" for the test framework


Stargator:
For the Dart exercises, we want to gradually introduce concepts related to the Dart language, test framework, and the built-in tools.

A lot of this is already done, by exposing users to problems written in Dart and asking for a solution. By working to understand the syntax of the Dart language, users will gradually learn more about Dart's features.

But if we keep with the simple setup for test suites from the first exercise, we risk not exposing other capabilities within the test framework. Like groups and the setup/teardown pattern.

So this issue will be just one step in that goal.

Judging by how the test suite for Gigasecond is laid out, I would recommend adding "setup" function to the test suite with comments explaining it's purpose and use. Also, add teardown as well.

There should be comments above the setup and teardown describing it's purpose in the simplest form.

For information about the test framework, [check out the link](https://pub.dartlang.org/packages/test).

Stargator:
@kabiir or @fwip

Once I get your first exercise PR merged in. Then this is the last issue left before we can judge if the track is ready for primetime. Let me know if either of you are interested or have questions.

raphalvessa:
I'm happy to contribute to this issue, I just don't see how the `setUp()` and `tearDown()` would help in this case. I don't see any duplicated code, or code that needs to be ran before and/or after each group test.

Stargator:
@raphalvessa By all means :joy_cat: take this on.

Well, the test suite would have to be reorganized. My thinking was to make some, if not all, the variables `birthDate`, `expectedDate`, `gigasecond` global. Then the setup/teardown could be used to set and re-set. I think the only one setup to handle would be `birthdate`, it would just have to be changed to the same value.

Then again, Gigasecond doesn't **have** to be be the exercise that is reworked for setup/teardown. If there's an exercise you think fits the needs better, go for it!

raphalvessa:
@Stargator Oh right, I thought you were referring exactly to the Leap exercise (though I just noticed that the title mentions the Gigasecond exercise). In that case, I'll look at all the current tests and see how I can improve them. :smiley:

Stargator:
@raphalvessa Not all the tests, just one test suite needs to introduce the setup/teardown

raphalvessa:
@Stargator Oh right, I'll find a good candidate and add it to that one only. Meanwhile, I found ways of optimizing tests that had repeated code (and not because a `setUp()` was missing). Would you like me to include those changes, or should I open an issue?

[...]

PR #55, opened 2017-10-07
Concept of setUp() for tests

raphalvessa:
The purpose of this PR is to introduce the concept of `setUp` for the test framework as raised in issue #31. I created global variables and used the `birthDate` and a fixed date. There is no `tearDown()` since Dart actively deletes objects (garbage collection).


exercises/gigasecond/test/gigasecond_test.dart
```diff
@@ -1,54 +1,60 @@
...
-
-      test("even works before 1970 (beginning of Unix epoch)", () {
-        expect(gigasecond.date(), equals(expectedDate));
-      });
-    });
+  group("Gigasecond", gigasecondTests);
+}
+
+void gigasecondTests() {
+  setUp(() {
```
Stargator: Since the exercises is not tied to the use of setup. Can you add a comment here to explain the purpose of setup?

Just for this first occurence

exercises/gigasecond/test/gigasecond_test.dart
```diff
@@ -1,39 +1,63 @@
...
-      gigasecond.date();
-      expect(gigasecond.date(), equals(expectedDate));
-    }, skip: true);
+  group("Gigasecond", gigasecondTests);
+}
+
+void gigasecondTests() {
+  // The setUp() function is used to share code between tests. Its callback (the
+  // anonymous function it takes as a parameter) is ran before every test in a
+  // group or test suite.
```
Stargator: I don't think "callback" should be used. May be too technical for new people to the language. Recommend to keep it simple.

I would recommend something like:
> // The setUp() function is used to share code between tests.
> // It is executed before every test in a  group or test suite.

Issue #57, opened 2017-10-07
Simplify Hello World test suite


raphalvessa:
The tests for the Hello World exercise in Dart test a `helloWorld.hello()` function that can be called with a parameter `name`. The description of the exercise doesn't mention any function parameters, and looking at the same exercise in other languages, none of them test for function parameters. I suggest sticking to the exercise description, and removing the tests that expect a `name`. After all, the Hello World should be the simplest, most basic exercise to introduce the language to new coders.

Stargator:
I see your point @raphalvessa.

I looked at the [specification for hello world](https://github.com/exercism/problem-specifications/blob/master/exercises/hello-world/canonical-data.json) and it doesn't have a case for parameters.

And I have looked at a handful of other tracks and could find a parameter in their hello world exercise. So I agree with you.

But I'll rope in our new maintainers @fwip @kabiir to see if they want to chime in.

raphalvessa:
I don't mind working on this, it's a very quick change.

Stargator:
Related PR merged in

PR #60, opened 2017-10-08
Simplify Hello World

raphalvessa:
Related to #57 

Removes the parameter `name` from the exercise and tests.


exercises/hello-world/lib/example.dart
```diff
@@ -1,3 +1,3 @@
 class HelloWorld {
-  String hello([String name = "World"]) => "Hello, $name!";
+  String hello() => "Hello, World!";
```
kabiir: I recommend changing the fat arrow (=>) to a block code. Beginners would want to know what exactly is happening here.

kabiir: Explicitly adding the return statement would prove helpful to new comers and make them feel familiar to the syntax if they come form a mainstream language such as c++, Java.

Stargator: @kabiir I take it you are in favor of this change which is being discussed in #57 ?

kabiir: Yeah, hello world is the kind of exercise which beginners should just be able to get past through it easily. They'll learn about parameters in proceeding exercises.

Traditionally *hello world* is about how to get output in a particular language. Which is using print() in dart but the test requires the output as a return and not as an actual output on the console. So this will do.

Stargator:
I disagree with the additional comments about the "fat arrow", 
We shouldn't have to comment and explain why we do anything. in the example file. Only other contributors or maintainers will see it.

raphalvessa:
@Stargator My bad, I'll get rid of it.

Stargator:
@raphalvessa Nothing to feel bad about. Just my opinion.

I appreciate you taking the time and effort to help us.

raphalvessa:
@Stargator My pleasure, I'm here to help and learn. 🙂

Stargator:
@kabiir I approve this, but not sure if you want to go over it again since you reviewed it too.

Stargator:
@raphalvessa You might want to wait on updating the branch until it's ready to merge.

PR #65, opened 2017-10-15
Revise Exercise READMEs using configlet generate command

Stargator:
Decided to run the `configlet generate` command on all the exercises and voila!

Most were missing a section on how to run tests. So included those updates here as well as added a README for raindrops which was apparently missing one.

### DIFF
--- a/exercises/anagram/README.md
+++ b/exercises/anagram/README.md
@@ -6,6 +6,16 @@ Given `"listen"` and a list of candidates like `"enlists" "google"
 "inlets" "banana"` the program should return a list containing
 `"inlets"`.
 
+
+To run the tests:
+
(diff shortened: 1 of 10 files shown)


kabiir:
What if some of the Readmes already have content form problem specification and contributors have added more? Is that possibility covered.

kabiir:
Did you manually update or actually used the configlet

kabiir:
The changes if automated by the configlet are quite impressive.

Stargator:
@kabiir I used the configlet generate command

kabiir:
Usually I'd say **NO** to bulk changes but in this case it's fine. Should proceed with merge.

PR #67, opened 2017-10-16
create-exercise can generate README.md

fwip:
This closes #62.

Not a lot of exciting things going on in this pull request. Generates `README.md` using `configlet generate`, and factors out a pattern for running processes.

Issue #61, opened 2017-10-14
Create scripts to to reduce commands used to validate Pull Requests


Stargator:
[Stemming from this comment](https://github.com/exercism/dart/pull/58#discussion_r144105347), the following commands can be run from a script:
1. `pub run test` - All the tests for Dart exercises can be run from the top level of the repo with this command
1. `bin/configlet lint .` - Checks the config.json for formatting issues
1. `bin/configlet fmt .` - Formats the config.json
1. `bin/check_formatting` - Checks All the Dart files for formatting issues
1. `pub run dart_style:format -l 120 -w .` - Formats all the Dart files

I recommend 2 and 4 can be in a script that just checks formatting of the project files.

Then 1, 3, and 5 can be put in a script that validates tests and automatically formats configlet and dart files.

Then there are two files that need to be updated with the new scripts:
* .travis.yml
* README.me

The scripts should be able to be run in Mac, Linux, and Windows with as little duplication as possible. Since we are using Dart, the scripts _could_ written in Dart, but it's up to the author.

kabiir:
Some ideas for the script :

- By default it'll run all 5 task for all exercises.
- There should be an option to perform those tasks for a specific exercise.
- It should have option to perform specific tasks for specific exercise. (maybe not, need opinions) 
- It's name should be sensible (e.g. validate-exercise) along with its args (--all, --nameOfExercise, etc) 
- Possibly hook the script to git-hooks (pre-push) a non-zero exit code will abort the push. 
- Should return zero exit code after successful completion, this will also be helpful in travis-ci
- more ideas here...

kabiir:
Few more ideas
- If canonical-data has tests, there should exists a `exercise_name_test.dart` file in `test` directory.
- Number of tests implemented `>=` Number of tests in canonical-data (It's fine if there are more).
- Each test implemented has flag `skip:true` except the first one.
- The actual `example.dart` should **not** be imported in the exercise_name_test.dart (common mistake).
- The `exercise_name.dart` should be imported in `exercise_name_test.dart`.
- If there exists a description.md in the problem-specification, there should be a readme.md (In case the contributor didn't use our create-exercise tool. )

Stargator:
> There should be an option to perform those tasks for a specific exercise.
That's possible for the scripts that check dart formatting, but the configlet commands check the entire repo's configuration and contents.

> It should have option to perform specific tasks for specific exercise. (maybe not, need opinions)
[...]

PR #70, opened 2017-10-24
Presubmit scripts

raphalvessa:
Adds presubmit scripts and scripts to check formatting to replace five current commands.


raphalvessa:
I suggest removing the current `bin/check_formatting` script and running `pub run dart_style:format -l 120 -n .` directly in the new `bin/check_formatting.dart`. I don't like the filename duplication, but I couldn't think of a better name. I'd like to hear your thoughts.

bin/presubmit.dart
```diff
@@ -0,0 +1,24 @@
...
+Future<Null> main() async {
+  final ProcessManager manager = new ProcessManager();
+  Process spawn;
+
+  print('Running test...');
+  spawn = await manager.spawn('pub', ['run', 'test']);
+  await spawn.exitCode;
+
+  print('Formatting config.json...');
+  spawn = await manager.spawn('bin/configlet', ['fmt', '.']);
```
kabiir: I'm not so sure about omitting configlet lint. Because there could be a human error. Maybe run the configlet lint _after_ formating to make sure it is infact a human error.

kabiir: Also make sure configlet actually exists before running it. Otherwise fetch it.

kabiir:
Maybe add an argument '--travis' so we can configure the behavior of the script when running on Travis. Since formatting dart files and the configlet can be done by the contributor we can skip these steps on Travis and only lint those files and do an early exit.

bin/presubmit.dart
```diff
@@ -0,0 +1,24 @@
...
+import 'dart:io';
+import 'package:io/io.dart';
+
+Future<Null> main() async {
+  final ProcessManager manager = new ProcessManager();
+  Process spawn;
+
+  print('Running test...');
+  spawn = await manager.spawn('pub', ['run', 'test']);
+  await spawn.exitCode;
```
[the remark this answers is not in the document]
raphalvessa: Good point. I was just following the order in the README file. Will change.

bin/presubmit.dart
raphalvessa: I'm not sure I understand. The command `bin/configlet lint .` is in the `check_formatting.dart` script. I separated the commands into two scripts as @Stargator suggested in #61.

kabiir: Oh, I didn't realize check formating was running before this script. My mistake 🙈. Although do we need to fetch the configlet twice? I'm not sure if the `before script` and `script` sections of Travis configuration  are isolated.

raphalvessa: That's ok. Well, `check_formatting.dart` and `presubmit.dart` are independent, so you can run them in whichever order you like, though I think it makes more sense to run `check_formatting.dart` first.

No, you only need to fetch the configlet once, but that's a good point. Since you're required to have a configlet to run these scripts, perhaps they could fetch it if one doesn't exist already.

I'm not sure about the Travis configuration. I believe they're isolated, but maybe @Stargator and @kytrinyx can enlighten us.

kabiir: Regardless of Travis configuration being isolated, both of these scripts will still need to check if configlet exists, and fetch it if not. Which is not that much code duplication but I think we can improve on this.

raphalvessa: I totally agree.

bin/check_formatting.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:async';
+import 'dart:io';
+import 'package:io/io.dart';
+
+Future<Null> main() async {
```
Stargator: main should return void.

bin/check_formatting.dart
```diff
@@ -0,0 +1,20 @@
...
+
+  print('Checking config.json formatting...');
+  spawn = await manager.spawn('bin/configlet', ['lint', '.']);
+  await spawn.exitCode;
+
+  print('Checking all Dart files formatting...');
+  spawn = await manager.spawn('bin/check_formatting', []);
+  await spawn.exitCode;
+
+  await sharedStdIn.terminate();
```
Stargator: The following lines can be reused:
* print
* spawn
* await

I recommend making a function that does at least the spawn and await. It can be called twice. Once for configlet and the second for check_formatting.

bin/presubmit.dart
```diff
@@ -0,0 +1,24 @@
...
+
+  print('Formatting all Dart files...');
+  spawn = await manager.spawn('pub', ['run', 'dart_style:format', '-l', '120', '-w', '.']);
+  await spawn.exitCode;
+
+  print('Running test...');
+  spawn = await manager.spawn('pub', ['run', 'test']);
+  await spawn.exitCode;
+
+  await sharedStdIn.terminate();
```
Stargator: Similar to the previous comment, some lines can be refactored to use a method that spawns and awaits.

pubspec.yaml
```diff
@@ -7,5 +7,6 @@ authors:
 dev_dependencies:
   args: '0.13.7'
   dart_style: '>=1.0.8 <2.0.0'
+  io: '0.3.0'
```
Stargator: I don't think io package is not needed. What it does we can do with Process.run or Process.start, I think,

Stargator:
I thought the intention was to have scripts that run before a user pushes or commits a change in our repo?

Stargator:
Oh sorry, one last thing. You need to add `bin/fetch-configlet` to check formatting and the pre-submit script. Because that will download the latest version of configlet into the bin/ directory.

bin/check_formatting.dart
raphalvessa: I followed the first code example for the `ProcessManager` from here: https://pub.dartlang.org/packages/io. `SharedStdIn.terminate()` returns a `Future<Null>`.

bin/check_formatting.dart
raphalvessa: I'm aware of the code repetition and I tried to do something about it without much success. Unfortunately I haven't had much time to improve this lately. I like your suggestion though, and will do that today.

pubspec.yaml
raphalvessa: The reason why I used `io` instead of `Process.run()` is that with the latter it's not possible to print the output while the command is running. For example, running the tests would display a blinking cursor for a few seconds, and then display the whole output at once. I then consulted with the Dart community, and one of the Dart devs from Google suggested using `io` which does the job perfectly.

raphalvessa:
@Stargator Yes, the intention is to run these scripts before the user pushes or submits a PR. I'm interested to know what made you thing that that wasn't the case.

I'm also fetching the configlet before running any of the scripts' commands. I'll push that later today.

bin/utils.dart
```diff
@@ -0,0 +1,33 @@
...
+
+    if (!configletFile.existsSync()) {
+      print('Fetching configlet...');
+      Process spawn = await manager.spawn('bin/fetch-configlet', []);
+      await spawn.exitCode;
+    }
+  }
+
+  /// Runs the [executable] with [arguments].
+  Future<Null> run(String executable, [List<String> arguments = const []]) async {
```
kabiir: Renaming `run` to `runCmd`  and changing the return type to `Future<int>` can also prove useful. Such as in our create-exercise script #67 we are using `Process.run`. The name `runCmd` clearly states its going to run a command, Once you rename that to `runCmd` we can reuse your code in other scripts too. Other scripts can call this function and if the `Future` resolves can do `something` and if it fails can do `something on fail`. They don't have to worry about the `exitCode` or outputting the process output, they get it for free. Similar to `fetchConfiglet` return type needs to be `Future<int>` so the `exitCode` can be utilized by the calling script. 
FYI spawn.exitCode is a `Future<int>`

bin/presubmit.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:async';
+import 'utils.dart';
+
+Future<Null> main() async {
+  final PresubmitUtils utils = new PresubmitUtils();
+
+  await utils.fetchConfiglet();
+
+  print('Formatting config.json...');
+  await utils.run('bin/configlet', ['fmt', '.']);
```
kabiir: How do we know that the `fetchConfiglet` was a success? Shouldn't there be a check before executing `configlet fmt .` ?

bin/check_formatting.dart
```diff
@@ -0,0 +1,17 @@
+import 'dart:async';
+import 'utils.dart';
+
+Future<Null> main() async {
+  final PresubmitUtils utils = new PresubmitUtils();
+
+  await utils.fetchConfiglet();
+
+  print('Checking config.json formatting...');
+  await utils.run('bin/configlet', ['lint', '.']);
```
kabiir: How do we know that the `fetchConfiglet` was a success? Shouldn't there be a check before executing `configlet fmt .` ?

bin/utils.dart
```diff
@@ -0,0 +1,33 @@
...
+    Process spawn = await manager.spawn(executable, arguments);
```
kabiir: Before you spawn the `executable` check whether or not it is in fact executable using `isExecutable`, If it's not  
```dart
print("The $executable is not executable, verify you have sufficient permissions to execute it.");
```
and `return false`.
You can checkout [this ](https://github.com/matanlurey/dart-future-examples/blob/master/bin/5_using_future_or.dart)example of `FutureOr` for this purpose.

bin/utils.dart
kabiir: You will need to use `FutureOr<int>` instead of `Future<int>` if you chose to add a check for `isExecutable`.

bin/utils.dart
kabiir: Windows doesn't have a concept of executables, but for other OSs we need to check this.

raphalvessa: I added the `isExecutable` check before running the command, but for some reason is always returns `false` when running `pub` commands. Probably because the user that runs the dart script on my system doesn't have permission to run `pub`, I don't know.

bin/utils.dart
```diff
@@ -0,0 +1,33 @@
...
+
+class PresubmitUtils {
+  ProcessManager manager;
+
+  PresubmitUtils() {
+    this.manager = new ProcessManager();
+  }
+
+  /// Fetches the configlet file if it doesn't exist already.
+  Future<Null> fetchConfiglet() async {
```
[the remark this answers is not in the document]
Stargator: @kabiir Recommend we document the usage either in the code or a markdown document. We don't really use our wiki section, which is why that wouldn't be my suggestion.

Stargator:
> @Stargator Yes, the intention is to run these scripts before the user pushes or submits a PR. I'm interested to know what made you thing that that wasn't the case.

@raphalvessa I was under the expectation we might by "presubmit" we were talking about configuring it so [git would automatically execute the script(s) when committing](https://git-scm.com/docs/githooks#_pre_commit). But that could have been a false understanding.

Cause I think that may have to be done locally, but I am not sure.

bin/utils.dart
```diff
@@ -0,0 +1,56 @@
+import 'dart:async';
```
Stargator: I suggest moving this outside of bin/ it's not an executable. Maybe a new directory lib/src or just lib/?

kabiir: lib/src IMO.

Stargator:
Just tagging this stems from #61

bin/check_formatting.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:async';
+import 'utils.dart';
+
+Future<Null> main() async {
+  final CommonUtils utils = new CommonUtils();
+
+  utils.fetchConfiglet().then((int exitCode) async {
+    if (exitCode == 0) {
+      print('Checking config.json formatting...');
+      await utils.runCmd('bin/configlet', ['lint', '.']);
```
Stargator: This should be updated to meet this command:
`configlet lint --track-id dart .` To avoid false positives about duplicate UUIDs

bin/check_formatting.dart
```diff
@@ -0,0 +1,19 @@
...
+Future<Null> main() async {
+  final CommonUtils utils = new CommonUtils();
+
+  utils.fetchConfiglet().then((int exitCode) async {
+    if (exitCode == 0) {
+      print('Checking config.json formatting...');
+      await utils.runCmd('bin/configlet', ['lint', '--track-id', 'dart', '.']);
+
+      print('Checking all Dart files formatting...');
+      await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '.']);
```
kabiir: Only perform configlet related commands inside the `fetchConfiglet().then` callback.

bin/check_formatting.dart
```diff
@@ -0,0 +1,19 @@
...
+
+      await utils.terminate();
+      print('Done!');
```
kabiir: Call `await utils.terminate();` and various output statements after the `fetchConfiglet().then` callback execution completes

bin/presubmit.dart
```diff
@@ -0,0 +1,22 @@
...
+      print('Formatting config.json...');
+      await utils.runCmd('bin/configlet', ['fmt', '.']);
+
+      print('Formatting all Dart files...');
+      await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-w', '.']);
+
+      print('Running test...');
+      await utils.runCmd('pub', ['run', 'test']);
+
+      await utils.terminate();
```
kabiir: Same as above

lib/src/utils.dart
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /// Terminates the global `stdin` listener.
+  Future<Null> terminate() async {
+    await ProcessManager.terminateStdIn();
+  }
+}
+
+/// Shows an error message with [exitCode].
+void _showError(exitCode) {
```
kabiir: As it is right now `_showError` is basically a wrapper for `print` statement.
You can return the `exitCode` here and conditionally print the error message
```dart
/// Shows an error message if [exitCode] is not 0.
int _exitCode(exitCode) {
    if (exitCode != 0 ) print ("Error...");
    return exitCode;
}
```
Doing so you won't need to repeat the `if` condition.

lib/src/utils.dart
```diff
@@ -0,0 +1,56 @@
...
+    //
+    // final result = isExecutable(executable);
+    // if (result is bool && !result || result is Future && !await result) {
+    //   print('Unable to run "$executable". Make sure that it\'s executable and that you have permission to run it.');
+    //   return new Future.value(1);
+    // }
+
+    Process spawn = await manager.spawn(executable, arguments);
+    if (await spawn.exitCode != 0) _showError(exitCode);
+    return spawn.exitCode;
```
kabiir: If you follow  the below suggestion for `_showError()` method, you can just 
```dart 
return _exitCode(spawn.exitCode));
``` 
and avoid duplicating `if` condition

bin/check_formatting.dart
Stargator: Ah okay, I didn't catch the use of async. I dunno why

exercises/hello-world/lib/example.dart
```diff
@@ -1,3 +1,5 @@
 class HelloWorld {
-  String hello([String name = "World"]) => "Hello, $name!";
+  String hello() {
+    return "Hello, World!";
+  }
```
Stargator: I don't think thus change is needed. The example.dart files are outside the scope of adding the presubmit script, in my opinion.

exercises/hello-world/lib/hello_world.dart
```diff
@@ -1,5 +1,5 @@
 class HelloWorld {
-  String hello([String name]) {
+  String hello() {
```
Stargator: Not within the scope of this work. If this is a bug or oversight from previous work, then I think the right step would be to create an issue or PR specificaly to resolve.

config.json
```diff
@@ -42,9 +43,9 @@
       "slug": "word-count",
       "topics": [
         "control_flow_loops",
-        "regular_expressions",
+        "lists",
         "maps",
-        "lists"
+        "regular_expressions"
```
Stargator: This ticket doesn't cover formatting the contents of other files. It's just to add the presubmit script.

exercises/hello-world/test/hello_world_test.dart
```diff
@@ -2,21 +2,7 @@ import "package:test/test.dart";
 import "package:hello_world/hello_world.dart";
 
 void main() {
-  var helloWorld = new HelloWorld();
-
-  test("should work", () {
-    helloWorld.hello();
+  test("returns \"Hello, World!\"", () {
+    expect(new HelloWorld().hello(), equals("Hello, World!"));
```
Stargator: Not within the scope of this work. If this is a bug or oversight from previous work, then I think the right step would be to create an issue or PR specificaly to resolve.

exercises/leap/test/leap_test.dart
```diff
@@ -10,7 +10,7 @@ void main() {
     });
 
     test("is introduced every 4 years to adjust about a day", () {
-      final year = new Year(2020);
+      final year = new Year(1996);
```
Stargator: This change was made in a separate PR that was already merged into master, no need to include it here.

exercises/phone-number/README.md
```diff
@@ -13,7 +13,7 @@ The format is usually represented as
 
 where `N` is any digit from 2 through 9 and `X` is any digit from 0 through 9.
 
-Your task is to clean up differently formated telephone numbers by removing punctuation and the country code (1) if present.
+Your task is to clean up differently formatted telephone numbers by removing punctuation and the country code (1) if present.
```
Stargator: This change was made in a separate PR that was already merged into master, no need to include it here.

test/exercises_test.dart
```diff
@@ -112,19 +112,19 @@ Running tests for: $packageName
...
+      final testPath = "${Directory.current.path}/exercises/$testName";
 
-      if (!await new Directory(testpath).exists()) {
-        throw new ArgumentError("No exercise with this name: $testname");
+      if (!await new Directory(testPath).exists()) {
+        throw new ArgumentError("No exercise with this name: $testName");
       }
 
-      await runTest(testpath);
+      await runTest(testPath);
```
Stargator: These changes were made in a separate PR that was already merged into master, no need to include it here.

tool/create-exercise
```diff
@@ -226,38 +242,32 @@ Future main(args) async {
...
 
-  if (!res.stderr.toString().isEmpty) {
-    stderr.write(res.stderr);
-  }
+  // Install deps
+  Directory.current = exerciseDir;
 
-  assert(res.exitCode == 0);
+  final pubSuccess = await runProcess("pub", ["get"]);
+  assert(pubSuccess);
```
Stargator: I believe these changes were made in a separate PR that was already merged into master, no need to include them here.

I recommend doing a rebase onto master to remove any changes introduced by previous commits to master.

lib/src/utils.dart
```diff
@@ -0,0 +1,56 @@
+import 'dart:async';
+import 'dart:io';
+import 'package:io/io.dart';
+
+class CommonUtils {
+  ProcessManager manager;
```
Stargator: I recommend making manager a private variable

tool/create-exercise
kabiir: Also  you don't necessarily need to update your local branch after making a PR, you can continue making commits to it and when the PR is ready to merge it can be update.

raphalvessa: Sorry guys, I messed up the commit history here. Rebasing is messing up everything, I'll be more careful next time.

Stargator: @raphalvessa No worries, if you want, I can rebase the branch for you.

raphalvessa: @Stargator Yes please, I don't seem to be able to get out of this mess. Thanks!

lib/src/utils.dart
```diff
@@ -0,0 +1,53 @@
...
+  /// If the [executable] isn't executable, returns exit code 1 and shows an
+  /// error message.
+  Future<int> runCmd(String executable, [List<String> arguments = const []]) async {
+    // Commented out because it always return false with "pub".
+    //
+    // final result = isExecutable(executable);
+    // if (result is bool && !result || result is Future && !await result) {
+    //   print('Unable to run "$executable". Make sure that it\'s executable and that you have permission to run it.');
+    //   return new Future.value(1);
+    // }
```
Stargator: If we are not using this, recommend removing the commented code

tool/create-exercise
Stargator: @raphalvessa Done.

lib/src/utils.dart
kabiir: I recommended implementing such, but it needs debugging. I'll see what I can do about that.

raphalvessa: @kabiir I think I know what's going on here. Anything under our `bin/` is executable, but all other commands aren't (that includes `pub` and all other system commands like `ls` and `pwd`). Looking again at the docs for `isExecutable()`, the first parameter is a path, and it returns `true` if this path contains a file that is executable. Well, commands like `pub` aren't a path, but rather an alias to a path, and I'm guessing that's why `isExecutable()` returns `false`. Since we have no way of checking if a command is a path or an executable, I think we should probably get rid of this check. I'd like to know what you think.

Stargator: Maybe keep it for the scripts, but use an alternative for commands like `pub` without it

raphalvessa: Ok, that sounds like a good idea. One function that has the `isExecutable()` check, and another that doesn't, and then it's up to the user which one they want to use. Thanks guys.

config.json
kabiir: Revert this change

config.json
```diff
@@ -45,7 +45,7 @@
         "control_flow_loops",
         "lists",
         "maps",
-        "regular_expressions"
+        "lists"
```
Stargator: Please fix this, it has control_flow_loops, lists, maps, and lists again. I would remove the second "lists"

Stargator: Recommend running `configlet fmt config.json` to format this file. You may need to run `fetch-configlet` to update your local version of configlet.

raphalvessa: Oops, I didn't notice the duplicate.

Stargator: @raphalvessa I would suggest the changes are reverted back to the original version:
> "control_flow_loops",
> "regular_expressions",
> "maps",
> "lists"

Stargator:
Unless otherwise contacted, @raphalvessa I'm going to rebase this branch and merge it in this weekend. Generally, it looks good to me.

PR #74, opened 2017-10-29
Fix commenting for Dart standard, rename variables

Stargator:
Just a bit of polishing up. The commenting didn't follow Dart standard of three slashes for documentation.

And a few variable names need to be updated. Constants follow lower camelcase like all variables.


fwip:
Looks good to me. 👍

PR #75, opened 2017-10-29
Add linting rules for errors

Stargator:
Revise the analysis-options.yaml file with some lint rules to prevent errors. Will be useful as the project grows.

PR #76, opened 2017-10-29
Address lint rule error for always_declare_return_types

Stargator:
The linting from #75 found the return type missing for these functions. Addressing it will also allow the test suite to match how the other test suites for exercises do this.


kabiir:
I will suggest you to wait for #75 and all required exercises are merged/updated for the launch and then re-lint.

Stargator:
All required exercises are merged in already. And I used #75 for this, but this PR is specifically for the rule always_declare_return_types.

So nothing is blocking this outside of approval.

PR #79, opened 2017-10-29
Minor changes to update docs for track page

Stargator:
@fwip @kabiir @raphalvessa I would appreciate some feedback on the docs cause I wrote most of them, if not all, and they could use some fresh eyes.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -11,6 +11,8 @@
         * [Instructions for Windows](https://www.computerhope.com/issues/ch000549.htm)
         * [Instructions for Linux](https://www.computerhope.com/issues/ch001647.htm)
         * [Instructions for Mac](https://www.pegaxchange.com/tag/macos-sierra-add-entry-to-path-variable/)
+
+#### Verify Installation
 * From a new command window or terminal, type `dart --version` to verify that Dart is now available. If the output shows the version number, then Dart is correctly installed on your computer.
 
 ## Dart Plugin for your IDE or Editor
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,13 +1,6 @@
 ## Dart
 * [Install Dart](https://www.dartlang.org/install)
-* Add the Dart SDK to your path.
 * From a new command window or terminal, type `dart --version` to verify that Dart is now available.
 
 ## Dart Plugin for your IDE or Editor
 * If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools) if it exists for your editor of choice.
-
-## Setting up Dart on your IDE
-* Setup the Dart SDK and Dartium by going in your IDE's settings and add the Dart SDK
-* Note: For HomeBrew Mac users the Dart SDK and Dartium locations are:
-  * SDK directory: HOMEBREW_INSTALL/dart/1.24.2/libexec
-  * Dartium: HOMEBREW_INSTALL/dart/1.24.2/Chromium.app
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,5 +1,14 @@
 ## Dart
-* [Install Dart](https://www.dartlang.org/install)
+* One option is to follow the [instructions for installation](https://www.dartlang.org/install) on dartlang.org.
+
+### Manual install instruction
+* Download a zip file from [this page](https://www.dartlang.org/install/archive).
+* Unzip the file into whatever directory you choose.
+* Add the full path of that directory to your PATH.
+    * Follow the links below for updating your PATH on Windows, Linux and Mac platforms:
+        * [Instructions for Windows](https://www.computerhope.com/issues/ch000549.htm)
+        * [Instructions for Linux](https://www.computerhope.com/issues/ch001647.htm)
+        * [Instructions for Mac](https://www.pegaxchange.com/tag/macos-sierra-add-entry-to-path-variable/)
 * From a new command window or terminal, type `dart --version` to verify that Dart is now available.
 
 ## Dart Plugin for your IDE or Editor
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,7 +1,9 @@
-## Dart
-* One option is to follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
+## Installing Dart
 
-### Manual install instruction
+### Automatic installation instructions
+* Follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
+
+### Manual install instructions
 * Download a zip file from [this page](https://www.dartlang.org/install/archive).
 * Unzip the file into whatever directory you choose.
 * Add the full path of that directory to your PATH.
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,5 +1,5 @@
 ## Dart
-* One option is to follow the [instructions for installation](https://www.dartlang.org/install) on dartlang.org.
+* One option is to follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
 
 ### Manual install instruction
 * Download a zip file from [this page](https://www.dartlang.org/install/archive).
@@ -9,7 +9,7 @@
         * [Instructions for Windows](https://www.computerhope.com/issues/ch000549.htm)
         * [Instructions for Linux](https://www.computerhope.com/issues/ch001647.htm)
         * [Instructions for Mac](https://www.pegaxchange.com/tag/macos-sierra-add-entry-to-path-variable/)
-* From a new command window or terminal, type `dart --version` to verify that Dart is now available.
+* From a new command window or terminal, type `dart --version` to verify that Dart is now available. If the output shows the version number, then Dart is correctly installed on your computer.
 
 ## Dart Plugin for your IDE or Editor
 * If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools) if it exists for your editor of choice.
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,9 +1,9 @@
 ## Installing Dart
 
-### Automatic installation instructions
+### Automatic installation instructions (recommended)
 * Follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
 
-### Manual install instructions
+### Manual install instructions (for advanced users)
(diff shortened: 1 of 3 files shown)


Stargator:
@raphalvessa Yeah, that note was for from my project's docs. I'll have to check where Dart is usually installed on windows. I use Mac for work and Ubuntu at home.

So as far as the Ubuntu install steps we'll know where dart is located, but other Linux flavors will use the zip file to extract it wherever the user wants.

I'll give that some thought.

docs/INSTALLATION.md
```diff
@@ -1,10 +1,10 @@
...
 * [Install Dart](https://www.dartlang.org/install)
 * Add the Dart SDK to your path.
-* From a new command window, type `dart --version` to verify that Dart is now available.
+* From a new command window or terminal, type `dart --version` to verify that Dart is now available.
 
 ## Dart Plugin for your IDE or Editor
-* If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools).
+* If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools) if it exists for your editor of choice.
 
 ## Setting up Dart on your IDE
```
kabiir: Is this section necessary?

We basically want to make sure they have dart installed and is present in PATH env variable. Configuring IDE or Editors should be either optional tips/hints section or completely omitted. Since anyone contributing must have some experience with dart  Even if its a student(completely new to dart) and wants to report a bug or help us improve something the previous section "Dart Plugin for..."  should suffice.

docs/INSTALLATION.md
```diff
@@ -1,10 +1,10 @@
 ## Dart
 * [Install Dart](https://www.dartlang.org/install)
 * Add the Dart SDK to your path.
-* From a new command window, type `dart --version` to verify that Dart is now available.
+* From a new command window or terminal, type `dart --version` to verify that Dart is now available.
```
kabiir: If they install dart by following steps described at `dart-lang.org/install` there shouldn't be any problem. But it's still possible that someone forgot to update there path variable or perhaps messed up. We should have steps to add dart to path variable.

docs/INSTALLATION.md
Stargator: Everything under docs will be shown on as the sections for exercism.io/languages/dart .

So these docs are meant for folks who could be new to Dart, but want to try out some exercises to experienced folks.

Some IDEs may not catch that you are using Dart and thus won't ask for the path to the Dart SDK. So I thought it was a safe addition.

Who knows, they could be new to programming in general and thus don't know what an IDE is. So maybe we could flesh that out?

docs/INSTALLATION.md
Stargator: Having steps to add Dart to the Path will need to tackle the Mac, Linux, and Windows versions.

I think mac will be the tricky part since it could have Dart via a tar.gz that was extracted or via HomeBrew (or other auto installation method)

Stargator:
@kabiir @raphalvessa I addressed your points, I think. Let me know what you think.

docs/INSTALLATION.md
```diff
@@ -1,13 +1,15 @@
 ## Dart
-* [Install Dart](https://www.dartlang.org/install)
-* Add the Dart SDK to your path.
-* From a new command window, type `dart --version` to verify that Dart is now available.
+* One option is to follow the [instructions for installation](https://www.dartlang.org/install) on dartlang.org.
```
raphalvessa: Just a minor thing, but instead of ending the sentence with "on dartlang.org" I'd say "on the Dart website", only because there's already a link to the installation page.

docs/INSTALLATION.md
```diff
@@ -1,13 +1,15 @@
...
-* If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools).
+### Manual install instruction
+* Download a zip file from [this page](https://www.dartlang.org/install/archive).
+* Unzip the file into whatever directory you choose.
+* Add the full path of that directory to your PATH.
+    * Follow the links below for updating your PATH on Windows, Linux and Mac platforms:
+        * [Instructions for Windows](https://www.computerhope.com/issues/ch000549.htm)
+        * [Instructions for Linux](https://www.computerhope.com/issues/ch001647.htm)
+        * [Instructions for Mac](https://www.pegaxchange.com/tag/macos-sierra-add-entry-to-path-variable/)
+* From a new command window or terminal, type `dart --version` to verify that Dart is now available.
```
raphalvessa: To make it clearer, and if you think it makes sense, maybe add another sentence like, "If the output shows the version number, then Dart is correctly installed on your computer."

docs/INSTALLATION.md
```diff
@@ -1,13 +1,15 @@
...
-* If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools).
+### Manual install instruction
+* Download a zip file from [this page](https://www.dartlang.org/install/archive).
+* Unzip the file into whatever directory you choose.
+* Add the full path of that directory to your PATH.
+    * Follow the links below for updating your PATH on Windows, Linux and Mac platforms:
+        * [Instructions for Windows](https://www.computerhope.com/issues/ch000549.htm)
+        * [Instructions for Linux](https://www.computerhope.com/issues/ch001647.htm)
+        * [Instructions for Mac](https://www.pegaxchange.com/tag/macos-sierra-add-entry-to-path-variable/)
+* From a new command window or terminal, type `dart --version` to verify that Dart is now available. If the output shows the version number, then Dart is correctly installed on your computer.
```
kabiir: Idk if it's required or if it'll help. 
Maybe add a tip explaining something like , 'in some cases the updated values for PATH variable don't take effect untill the terminal is restarted/refreshed or a new one is stared. It's better to start a new session.' This will go to the exercism website dart page right?

docs/INSTALLATION.md
```diff
@@ -1,13 +1,15 @@
 ## Dart
-* [Install Dart](https://www.dartlang.org/install)
-* Add the Dart SDK to your path.
-* From a new command window, type `dart --version` to verify that Dart is now available.
+* One option is to follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
 
-## Dart Plugin for your IDE or Editor
-* If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools).
+### Manual install instruction
```
kabiir: If there is a 'Manual installation instructions' there should be an 'Automatic installation instructions' too.

docs/INSTALLATION.md
```diff
@@ -1,13 +1,15 @@
 ## Dart
-* [Install Dart](https://www.dartlang.org/install)
-* Add the Dart SDK to your path.
-* From a new command window, type `dart --version` to verify that Dart is now available.
+* One option is to follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
```
kabiir: So the section title is 'Dart'? I don't understand.

kabiir: I also suggest adding a table of contents.

docs/INSTALLATION.md
kabiir: The links provided for all 3 platforms might or might not contain this tip but since we are providing an instruction `dart --version`  for confirmation of PATH variable it should have this tip.

raphalvessa: That's a valid point, but at least the Linux and macOS instructions already cover this.

docs/INSTALLATION.md
fwip: Agreed, I suggest changing the title to "Installing Dart" or "Recommended installation instructions"

Stargator: @fwip That's what I have done locally. Will be addressing this feedback soon.

docs/INSTALLATION.md
Stargator: Those instructions are on the dart-lang website. Since we link to it, we don't need to duplicate the steps here otherwise, they may change in the future depending on the platform.

docs/INSTALLATION.md
Stargator: I though this was covered by stating "From a new command window or terminal ..."

docs/INSTALLATION.md
kabiir: No, I was talking about the section titles, so if there's a manual installation section there should be an automatic installation section too.

kabiir: Consider the following structure :

# Installation   


### Automatic   


###  Manual

docs/INSTALLATION.md
kabiir: @Stargator Yes it does. And it's clear too. But English is not my native language and this is one of the many mistakes I used to make when I was starting out. Even though the instruction says to start a new terminal I used to continue using the same one. Also note that this instruction is in continuation with 3 links for 3 platforms, a student might just use the same terminal thinking its a continuous process.

So I suggest to add a section called "Verifying dart installation". This verifying of installation is helpful in both automatic and manual installations.   

The significant word in our instruction is '**new** ' it's very small and can be easily missed. So another suggestion is instead of adding a new section maybe just highlight the word to emphasize its importance. 

What do you think?

kabiir: It might very well be an extreme case and never happen. This is why I chose to comment instead of requesting changes.

Stargator: Thanks @kabiir for clarifying. That's a good perspective I hadn't considered. I separated the verify step into another section.

Stargator:
@kabiir @raphalvessa Ready for another round.

docs/INSTALLATION.md
```diff
@@ -1,13 +1,19 @@
-## Dart
-* [Install Dart](https://www.dartlang.org/install)
-* Add the Dart SDK to your path.
-* From a new command window, type `dart --version` to verify that Dart is now available.
+## Installing Dart
 
-## Dart Plugin for your IDE or Editor
-* If you are using an IDE or an editor, [install and configure the Dart plugin](https://www.dartlang.org/tools).
+### Automatic installation instructions
```
kabiir: Suggestion : add `(recommended)` in section title.

docs/INSTALLATION.md
```diff
@@ -1,13 +1,19 @@
...
+* Follow the [instructions for installation on the Dart website](https://www.dartlang.org/install).
+
+### Manual install instructions
```
kabiir: Suggestion : add `(for advanced users)`  in section title.

Stargator:
@raphalvessa and @kabiir Addressed all suggestions

Stargator:
I'll rebase this tonight to account for recent changes.

PR #80, opened 2017-11-12
Fix invalid exercise UUID

nywilken:
The previous version of `configlet uuid` was known to generate invalid UUIDs. The latest release of Configlet [v3.6.1](https://github.com/exercism/configlet/releases) fixes that issue for newly added exercises, but existing exercise UUIDs need to be updated manually.

This change pertains to exercism/configlet#99

### DIFF
--- a/config.json
+++ b/config.json
@@ -111,7 +111,7 @@
         "strings"
       ],
       "unlocked_by": "gigasecond",
-      "uuid": "d6657d4a-0257-0c80-6667-fddd127a571f0ad1fd4"
+      "uuid": "3d4e326d-2419-4e93-ac29-5727dbc3fbf1"
     },
     {
       "core": true,


Stargator:
@nywilken Most of the current UUIDs I believe were generated without configlet, but an external tool. Form reading issue 99, I am not sure exactly why the existing UUIDs need to change. Shouldn't the lint tool catch any clashes?

Stargator:
Also is there a way to check for invalid UUIDs and only change those?

nywilken:
@Stargator thanks for looking into the PR. There is actually just one invalid UUID that I am replacing in this PR "d6657d4a-0257-0c80-6667-fddd127a571f0ad1fd4". Looking at the format 8-4-4-4-19 it was most likely generated by the old version of Configlet. 

>  I am not sure exactly why the existing UUIDs need to change. Shouldn't the lint tool catch any clashes?

The lint command checks for UUID clashes within tracks, but it does not actually check that the UUID is indeed a valid UUID (https://en.wikipedia.org/wiki/Universally_unique_identifier#Standards).  

I used a WIP build of Configlet (https://github.com/exercism/configlet/pull/107) to fix the UUID. It checks a track's config.json file for any invalid UUIDs and replaces them with a new UUID if the `--repair` flag is provided at runtime. So only one UUID change is necessary at this time.

```bash
› configlet uuid doctor . --repair
[X] d6657d4a-0257-0c80-6667-fddd127a571f0ad1fd4 (dart:raindrops)                                                               
```

Please let me know if that answers your question.

PR #86, opened 2017-12-07
New Exercise: Acronym

Stargator:
Implement [specification from problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/acronym)

@exercism/dart


Stargator:
The specification for this exercise was updated to fit a new definition, double check that there aren;t any additional changes

exercises/acronym/lib/example.dart
```diff
@@ -0,0 +1,15 @@
+class Acronym {
+  String abbreviate(final String phrase) {
+    StringBuffer acronymBuilder = new StringBuffer();
+
+    RegExp expression = new RegExp(r'(^[A-z])|(\s[A-z])|-[A-z]');
+
+    List<Match> matchList = expression.allMatches(phrase).toList();
+
+    for (Match match in matchList) {
+      acronymBuilder.write(match.group(0).trim().toUpperCase().replaceAll("-", ""));
```
jvarness: If we were able to use the split to get rid of the characters that aren't alpha characters, the `replaceAll` may no longer be necessary.

exercises/acronym/lib/example.dart
```diff
@@ -0,0 +1,15 @@
+class Acronym {
+  String abbreviate(final String phrase) {
+    StringBuffer acronymBuilder = new StringBuffer();
+
+    RegExp expression = new RegExp(r'(^[A-z])|(\s[A-z])|-[A-z]');
```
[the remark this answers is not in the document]
Stargator: Sure I'll look at that.

config.json
```diff
@@ -108,11 +108,25 @@
...
+        "equality",
+        "regular_expressions",
+        "strings"
+      ]
+    },
     {
       "slug": "phone-number",
       "uuid": "8649b2b4-02f6-4947-92b9-d1cc0d7c3594",
       "core": false,
-      "unlocked_by": "word-count",
```
jvarness: Does this mean a new exercise needs to be unlocked by word-count other than phone-number?

Stargator: Core exercises **cannot** be unlocked by another, so the order of core exercises matters as far as how the user will progressed through them,

Non-core exercises, I believe, **have to** be unlocked by another exercise. The `configlet` tool will check for anything that doesn't meet these conditions.

exercises/acronym/lib/example.dart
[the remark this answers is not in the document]
Stargator: I'm having trouble with `.split`. I have the regular expression correct (tested it), but split isn't working as I expected.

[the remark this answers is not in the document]
Stargator: Got it, it wasn't the expression, it was how I was creating an instance of RegExp...I wasn't

Stargator:
Hey @jvarness It's ready for a another review

exercises/acronym/lib/example.dart
```diff
@@ -0,0 +1,15 @@
+class Acronym {
+  String abbreviate(final String input) {
+    StringBuffer acronymBuilder = new StringBuffer();
+
+    RegExp separatorsExp = new RegExp(r'( )|(-)');
```
jvarness: So now we're only splitting on spaces and hyphens?

Should we split on any kind of white space? I think `\w` matches all whitespace.

We don't need to consider any other non-alphanumeric characters as part of this?

jvarness: I guess that makes a lot of sense based on the test cases. We don't have to split on all non-alphanumeric characters.

Although, I'm curious... If any splitted substring started with a non-alphanumeric character... We could see some really weird acronyms.

Stargator: @jvarness If you can find an acronym that starts with that (phrase, company name, etc), then we can certainly add that as a test case. :)

PR #87, opened 2017-12-07
New Exercise: Armstrong numbers

Stargator:
Add new exercise based on [this specification](https://github.com/exercism/problem-specifications/tree/master/exercises/armstrong-numbers)

@exercism/dart

### DIFF
--- a/exercises/armstrong-numbers/lib/example.dart
+++ b/exercises/armstrong-numbers/lib/example.dart
@@ -1,0 +1,20 @@
+import 'dart:math' show pow;
+
+class ArmstrongNumbers {
+
+  /// The parameters and variables within the method that are set to final in order to prevent the accidentally modification of the value.
+  /// As those variables are not needed to be changed.
+  bool isArmstrongNumber(final int number) {
+    final String numberAsString = number.toString();
+    final int numOfDigits = numberAsString.length;
+    int sum = 0;
+
+    for (int count = 0; count < numOfDigits; count++) {
+      final String digitAsString = numberAsString.substring(count, count + 1);
+      final int digit = int.parse(digitAsString);
+      sum += pow(digit, numOfDigits);
+    }
+
+    return number == sum;
+  }
+}
--- a/config.json
+++ b/config.json
@@ -28,6 +28,19 @@
     },
     {
       "core": true,
+      "difficulty": 1,
+      "slug": "armstrong-numbers",
+      "topics": [
+        "equality",
+        "integers",
+        "logic",
+        "strings"
+      ],
+      "unlocked_by": null,
+      "uuid": "4f650701-9568-4d95-9f72-be4bde2c90ec"
+    },
+    {
+      "core": true,
       "difficulty": 2,
       "slug": "difference-of-squares",
       "topics": [
--- a/exercises/armstrong-numbers/README.md
+++ b/exercises/armstrong-numbers/README.md
@@ -1,0 +1,29 @@
+# Armstrong Numbers
+
+An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a number that is the sum of its own digits each raised to the power of the number of digits.
+
+For example:
+
+- 9 is an Armstrong number, because `9 = 9^1 = 9`
+- 10 is *not* an Armstrong number, because `10 != 1^2 + 0^2 = 2`
+- 153 is an Armstrong number, because: `153 = 1^3 + 5^3 + 3^3 = 1 + 125 + 27 = 153`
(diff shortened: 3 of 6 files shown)


exercises/armstrong-numbers/lib/example.dart
```diff
@@ -0,0 +1,17 @@
+import 'dart:math' show pow;
+
+class ArmstrongNumbers {
+  bool isArmstrongNumber(final int number) {
+    final String numberAsString = number.toString();
+    final int numOfDigits = numberAsString.length;
+    int sum = 0;
+
+    for(int count = 0; count < numOfDigits; count++) {
```
kabiir: Suggestion : introduce `foreach`

exercises/armstrong-numbers/lib/example.dart
```diff
@@ -0,0 +1,17 @@
+import 'dart:math' show pow;
```
kabiir: This is a good place to introduce lib/module imports.
Add comment explaining this is a selective import which only imports `pow` however this is not required. 
For a novice student selective import might seem too much and might not understand its use and potential but for experienced devs it's probably a known concept.

exercises/armstrong-numbers/lib/example.dart
```diff
@@ -0,0 +1,17 @@
...
+
+class ArmstrongNumbers {
+  bool isArmstrongNumber(final int number) {
```
kabiir: Suggestion: Add comment for reasoning usage of `final`

exercises/armstrong-numbers/lib/example.dart
```diff
@@ -0,0 +1,17 @@
...
+    final String numberAsString = number.toString();
+    final int numOfDigits = numberAsString.length;
```
kabiir: This could've easily been a const , add comment reasoning choice of final

kabiir: Add early exists for `len = 1; return true` and `len = 2; return false` because novice students need to learn early exits are a good idea and help on optimizations too.

exercises/armstrong-numbers/lib/example.dart
Stargator: The example solution shouldn't be readied seen by the user. So it wouldn't matter whether it uses foreach or not.

Though, we could add a suggestion to the README. Maybe a link to an example on dartlang.org?

exercises/armstrong-numbers/lib/example.dart
Stargator: Most students may not think to look for the repository of the track let alone the solution file.

Though they may want to look at out other people solved the problem via the website. I don't see the value in updating the example as it's not a primary resource for a given user.

exercises/armstrong-numbers/lib/example.dart
Stargator: I disagree about early exists. I prefer one entry, one exit. We are not here to teach people the language, but provide them with a problem and then resources to utilize in their search for a solution.

The example will not be a primary resource for them. Unless they are already familiar with github, they may not consider looking at the track's own solution to a problem.

No where in our documentation do we suggest or promise that the solutions given for an exercise meet a criteria. If you want to have that discussion, then we should open an issue to discuss it.

But I am not going to do more work than we've outlined in our documentation.

kabiir: >No where in our documentation do we suggest or promise that the solutions given for an exercise meet a criteria.

Yes, but I was under the impression that we're supposed introduce better ways to do things in an effort to make students better at dart.

exercises/armstrong-numbers/lib/example.dart
kabiir: Should we add a discoverable way for students to checkout example.dart?

kabiir: Reference links to dartlang website for syntax/language specific hints is a good idea.

exercises/armstrong-numbers/lib/example.dart
Stargator: No, we do not introduce "better ways" of doing things. The user is never required to follow someone else's footsteps in how a problem is resolved. They would never see the example solution provided unless they seek out the GitHub repo.

We may suggest Dart concepts to use in the exercise's README, but not in the example solution.

Stargator:
The specification for this exercise was updated to fit a new definition, double check that there aren;t any additional changes

Stargator:
The specification for this exercise was updated to fit a new definition, double check that there aren;t any additional changes

exercises/armstrong-numbers/lib/example.dart
Stargator: > This could've easily been a const , add comment reasoning choice of final

So far, I don't see it as being easy. But I'll leave a comment otherwise.

PR #99, opened 2018-01-30
Launch Track

Stargator:
@exercism/dart With all the issues in [Polish Up Milestone](https://github.com/exercism/dart/milestone/2), it seems Launch #1 is ready to go!

Upon merging of this PR, the Dart Track will no longer be under the Upcoming section, but will be listed as a normal ready for use track.

Once this PR is merged, I'll also close the Polish Up milestone. I think we have a few PRs listed as low-priority that we can move up. I'd like get all the new exercises done and wrap up the presubmit-scripts.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,5 +1,5 @@
 {
-  "active": false,
+  "active": true,
   "exercises": [
     {
       "core": true,


Stargator:
Okay it's almost been a week. I'm going ahead and activating the track.

PR #115, opened 2018-04-06
leap: add more test cases

jvarness:
I added more test cases to account for dates less than 400 years old. I also wasn't sure why `skip: true` was being set on every test so I removed those.

### DIFF
--- a/exercises/leap/test/leap_test.dart
+++ b/exercises/leap/test/leap_test.dart
@@ -24,5 +24,20 @@ void main() {
       final bool result = leap.leapYear(2000);
       expect(result, equals(true));
     }, skip: true);
+
+    test("is introduced every 4 years to adjust about a day before 400 A.D.", () {
+      final bool result = leap.leapYear(4);
+      expect(result, equals(true));
+    }, skip: true);
+
+    test("is skipped every 100 years to remove an extra day before 400 A.D.", () {
+      final bool result = leap.leapYear(300);
+      expect(result, equals(false));
+    }, skip: true);
+
+    test("is reintroduced every 400 years to adjust another day including 400 A.D.", () {
+      final bool result = leap.leapYear(400);
+      expect(result, equals(true));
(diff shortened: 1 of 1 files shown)


jvarness:
The first iteration of my solution for leap failed when I introduced these tests. I suspect a lot of folks might also have incorrect solutions.

Stargator:
The `skip` is used for all but the first test. When users first run the tests we want them to tackle the simpler cases first then progress to increasingly more complex cases.

Did you verify the example.dart still works for the new test cases?

jvarness:
I did, they all pass with the example!

Just out of curiosity: is the skipped tests a pretty common pattern with other exercism projects as well?

jvarness:
@Stargator I've rebased these tests with the latest from `master`. I also put the `skip` flag on them per your request. Let me know if you have further questions.

Stargator:
> Just out of curiosity: is the skipped tests a pretty common pattern with other exercism projects as well?

If the language or framework supports it, I would say yes.

exercises/leap/test/leap_test.dart
```diff
@@ -24,5 +24,23 @@ void main() {
...
+    }, skip: true);
+
+    test("is skipped every 100 years to remove an extra day before 400 A.D.", () {
+      final bool result = leap.leapYear(300);
+
+      expect(result, equals(false));
+    }, skip: true);
+
+    test("is reintroduced every 400 years to adjust another day before 400 A.D.", () {
+      final bool result = leap.leapYear(400);
```
Stargator: @jvarness Not sure I understand this, how is the year "400" "before 400 A.D."?

jvarness: It's not lol, I'll fix this.

Stargator:
Okay, just rebase onto the latest of master and squash the two commits into a single commit with this message: `leap: add more test cases`. Then I'll approve it

PR #121, opened 2018-04-20
bob: Add requirement to README to match tests

camuthig:


Stargator:
@camuthig It seems you need to rebase your branch onto master as there were some changes.

Issue #120, opened 2018-04-18
travis.yml: Remove presubmit from config and ensure builds fail when tests fail.


Stargator:
During the Scrabble Score Pull Request, it was revealed that the Travis builds were passing even when files needed to be formatted. Additionally, the presubmit script was being executed, but that only needs to happen before submitting a PR, not when a PR is being built.

Additionally, when tests either fail or time out the Travis Build still passes when it should fail.

jvarness:
`dartfmt` has an option to do a dry run and set the exit code to 1 if there are changes. That might be a good option.

Stargator:
@jvarness We ran it via the check_formatting scripting, so it runs on all platforms.

jvarness:
@Stargator if your environment has the dart sdk installed though, won't it have `dartfmt` installed by default? Are there installations that don't include `dartfmt`?

Stargator:
@jvarness You are right, the script allows checks the formatting of the config.json. So instead of a single line for each command we rolled similar commands together in the check_formatting.dart and the presubmit.dart scripts.

Stargator:
> @Stargator if your environment has the dart sdk installed though, won't it have dartfmt installed by default? Are there installations that don't include dartfmt?

An additional point to make is that the SDK has a set version of the dartfmt package (dart_style). So if you want a newer version you'd have to run `pub global activate dart_style`. So if we add that to the script, then it would more justify manually running the dartfmt executable since it would follow updating the global package version.

jvarness:
@Stargator I've been working on this but I've hit some snags.

Looks like the Dart runner for Travis doesn't honor when exit codes are set within a Dart script. I'll need to investigate a little more and get back to you.

I figured out how to make formatting fail and how to run all the tests, but there's weird stuff happening with setting exit codes.

Stargator:
I've experienced this with Jenkins. I resolved it by throwing an error when the exit code was non-zero.

jvarness:
Throwing the error is fine, but after rebasing with master, it looks like the jobs are now stalling when the formatting of all the files is checked:

https://travis-ci.org/jvarness/dart

Stargator:
Resolved in #129

PR #129, opened 2018-07-09
#120 ensure failures for formatting and tests

jvarness:
Going to see if a PR into exercism's repo will work better than mine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,10 @@ sudo: false
 dart:
   - stable
 
-dart_task:
-  - test: --platform vm
-
 cache:
   directories:
     - $HOME/.pub-cache
 
-before_script:
-  - pub run bin/check_formatting.dart
-
 script:
-  - pub run bin/presubmit.dart
+  - pub run bin/check_formatting.dart
+  - pub run test
--- a/bin/check_formatting.dart
+++ b/bin/check_formatting.dart
@@ -1,17 +1,29 @@
 import 'dart:async';
 import 'package:exercism_dart/src/utils.dart';
 
-Future<Null> main() async {
+Future main() async {
   final CommonUtils utils = new CommonUtils();
+  int errorCode = 0;
 
   if (utils.fetchConfiglet() == 0) {
     print('Checking config.json formatting...');
-    await utils.runCmdIfExecutable('bin/configlet', ['lint', '--track-id', 'dart', '.']);
+    errorCode = await utils.runCmdIfExecutable('bin/configlet', ['lint', '--track-id', 'dart', '.']);
+
+    if (errorCode != 0) {
+      print('Checking config.json formatting failed!!');
+      await utils.terminate();
+      throw new StateError('Formatting failed.');
+    }
   }
 
   print('Checking all Dart files formatting...');
-  await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '.']);
+  errorCode = await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '--set-exit-if-changed', '.']);
+
+  if (errorCode != 0) {
+    print('Checking exercise formatting failed!!');
+    await utils.terminate();
+    throw new StateError('Formatting failed.');
+  }
 
   await utils.terminate();
-  print('Done!');
 }
--- a/test/exercises_test.dart
+++ b/test/exercises_test.dart
@@ -1,5 +1,3 @@
-@Timeout(const Duration(seconds: 120))
-
 import "dart:io";
 import "dart:async";
 
@@ -68,7 +66,6 @@ Running tests for: $packageName
   await runCmd(["cp", "lib/${packageName}.dart", "lib/${packageName}.dart.bu"]);
   await runCmd(["cp", "test/${packageName}_test.dart", "test/${packageName}_test.dart.bu"]);
   try {
-
     String inPlaceOption = Platform.isMacOS ? "--in-place" : "-i";
 
     for (List<String> cmds in [
@@ -129,5 +126,5 @@ void main() {
 
       await runTest(testPath);
     }
-  });
+  }, timeout: new Timeout(new Duration(seconds: 120)));
 }


test/exercises_test.dart
```diff
@@ -1,5 +1,3 @@
-@Timeout(const Duration(seconds: 120))
```
jvarness: I removed this because it was causing the tests to time out every single time we ran this set.

Now, when people add new tests, we won't need to remember to update this.

jvarness:
Fixes #120

jvarness:
@Stargator please review. Many of the changes here are formatting changes from running `dartfmt`  on the files. I recommend that we switch to using `dartfmt` since it's the built-in formatter for dart.

bin/check_formatting.dart
```diff
@@ -1,17 +1,27 @@
...
   print('Checking all Dart files formatting...');
-  await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '.']);
+  errorCode = await utils.runCmd('dartfmt', ['-n', './']);
+  if (errorCode != 0) {
+    print('Checking exercise formatting failed!!');
+    await utils.terminate();
+    throw new StateError('Formatting failed.');
+  }
 
   await utils.terminate();
```
jvarness: The build was stalling because I forgot to add this terminator at the end.

test/exercises_test.dart
Stargator: Without it, we were getting timeouts, so I'm interested in the impact.

Stargator:
@jvarness The formatting changes should be in a separate PR. Since those changes would result in a failed build, they need to be merged in before this PR.

.travis.yml
```diff
@@ -5,15 +5,10 @@ sudo: false
 dart:
   - stable
 
-dart_task:
-  - test: --platform vm
```
Stargator: Not familiar with Travis's Dart support, what's the impact of removing this?

bin/check_formatting.dart
```diff
@@ -1,17 +1,27 @@
 import 'dart:async';
 import 'package:exercism_dart/src/utils.dart';
 
-Future<Null> main() async {
+Future main() async {
   final CommonUtils utils = new CommonUtils();
-
+  int errorCode = 0;
```
Stargator: Add a  line between this and the if statement.

bin/check_formatting.dart
```diff
@@ -1,17 +1,27 @@
...
   if (utils.fetchConfiglet() == 0) {
     print('Checking config.json formatting...');
-    await utils.runCmdIfExecutable('bin/configlet', ['lint', '--track-id', 'dart', '.']);
+    errorCode = await utils.runCmdIfExecutable(
+        'bin/configlet', ['lint', '--track-id', 'dart', '.']);
+    if (errorCode != 0) {
```
Stargator: Add a line above the if statement.

bin/check_formatting.dart
```diff
@@ -1,17 +1,27 @@
...
+      print('Checking config.json formatting failed!!');
+      await utils.terminate();
+      throw new StateError('Formatting failed.');
+    }
   }
 
   print('Checking all Dart files formatting...');
-  await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '.']);
+  errorCode = await utils.runCmd('dartfmt', ['-n', './']);
+  if (errorCode != 0) {
```
Stargator: Add a line above the if statement.

jvarness:
@Stargator if I don't format the code then the build for this PR will fail. Is that ok?

.travis.yml
jvarness: So this was a little weird:

We used this task to run the tests, but then we ran presubmit prior that would execute `pub run test`. Instead, I've changed this to run the formatting check, then I run the tests afterwords. I think we could potentially remove the presubmit.dart script after this.

I think we could avoid specifying the platform since we're just running everything within the Dart VM.

Stargator:
@jvarness That's okay, because there will be a PR with the formatting fixes that would be merged in first.

When you do the formatting do you run (essentially) `dartfmt -i 0 -l 120 -w .`?

jvarness:
I'm running `dartfmt -n .` which does a dry run and sets a non-0 exit code if a file has changed. I'm omitting `-w` on purpose since in CI it doesn't matter if we overwrite the files. The default for `-i` is 0, but the default for `-l` is actually 80.

I can update [this](https://github.com/exercism/dart/pull/129/files#diff-19d9eac277b3ea26e1d3fbfa63a0c5b3R21) to allow us to have a default line length of 120 if that's what we're considering our standard.

jvarness:
Correction: `-n` only does a dry run. `--set-exit-if-changed` sets an exit code if it's not changed. In this case, I'm using both options.

jvarness:
I guess what we'd need to do is make it so that the presubmit and the check_formatting script match up with the formatting. Otherwise, presubmit will make check_formatting fail every time.

Should I make the check_formatting script match the presubmit script?

bin/check_formatting.dart
```diff
@@ -1,17 +1,30 @@
...
+    if (errorCode != 0) {
+      print('Checking config.json formatting failed!!');
+      await utils.terminate();
+      throw new StateError('Formatting failed.');
+    }
   }
 
   print('Checking all Dart files formatting...');
-  await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '.']);
+  errorCode = await utils.runCmd('dartfmt', ['-n', '--set-exit-if-changed', './']);
```
Stargator: The formatting command should still include `-l 120`.

Stargator:
I was just asking when you made the formatting changes to the branch what were the options you used.

> Should I make the check_formatting script match the presubmit script?

This is how I think they should look based on what you posted:
**check_formatting**
`await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-n', '--set-exit-if-changed', '.']);`

**presubmit**
`await utils.runCmd('pub', ['run', 'dart_style:format', '-l', '120', '-w', '.']);`

jvarness:
@Stargator went ahead and changed the commands to include the max line length of 120, but I kept using `dartfmt` rather than `pub run dart:style`.

jvarness:
@Stargator are we good on this one?

jvarness:
@Stargator I think I've addressed everything on here.

Stargator:
@jvarness My previous fix related to macOs users was not properly formatted. So it caused this to fail.

I pushed a fix to this branch and merged it with your commit. It did fix the failure in the previous travis build.

Stargator:
@jvarness WOOT! WOOT! 💯 

This is going to be a BIG help! Thanks again.

PR #133, opened 2018-07-23
Compensate when tests are executed on MacOS

Stargator:
I don't have a Windows machine to run the tests on, so I cannot verify that `pub run test` works in Windows.

I have noticed that MacOS was producing files with "-e" or "e" appended to the extension, so I resolved that.


jvarness:
Looks good to me!

PR #134, opened 2018-07-23
Enable Support for Dart 2

Stargator:
This enables support for Dart 2.0 from the top-level tests and scripts to the exercises.

Dart 1.0 is still supported, so users should still be able to run tests in either version of the SDK.

We may decide at a later time when to drop all support for Dart 1.0.

@jvarness If you could take a look when you have some time, I'd appreciate it.


jvarness:
Hey @Stargator, I'll see if I can't review this by the end of the week.

exercises/armstrong-numbers/lib/example.dart
```diff
@@ -6,7 +6,7 @@ class ArmstrongNumbers {
   bool isArmstrongNumber(final int number) {
     final String numberAsString = number.toString();
     final int numOfDigits = numberAsString.length;
-    int sum = 0;
+    num sum = 0;
```
jvarness: Did this have to change because pow returns num? Does the fact that we're comparing `sum` to `number` and the two are different data types?

exercises/gigasecond/test/gigasecond_test.dart
```diff
@@ -1,47 +1,48 @@
...
 
 void main() {
-  group('Add one gigasecond to the input.', gigasecondTests);
+  group("Gigasecond", gigasecondTests);
 }
 
 void gigasecondTests() {
   test("date only specification of time", () {
-    final DateTime birthDate = new DateTime.utc(2011, DateTime.APRIL, 25);
+    final DateTime birthDate = new DateTime.utc(2011, polyfill.DateTime.april, 25);
```
jvarness: If we're using the Dart 2 SDK, do we really need to use these polyfills? Or is it because you specified `>=1.24.0` in the version?

test/exercises_test.dart
```diff
@@ -76,7 +78,7 @@ Running tests for: $packageName
       ["sed", "-i", "-e", "s/\\bskip:\\s*true\\b/skip: false/g", "test/${packageName}_test.dart"],
 
       /// Pull dependencies
-      ["pub", "get"],
+      ["pub", "upgrade"],
```
jvarness: Why `upgrade` instead of `get`? Couldn't that cause unforeseen issues during CI?

exercises/phone-number/lib/example.dart
```diff
@@ -29,7 +29,7 @@ class PhoneNumber {
       onlyDigits = onlyDigits.substring(1);
     }
 
-    if (onlyDigits.length < 10 || onlyDigits.length > 10) return null;
+    if (onlyDigits.length > 10 || onlyDigits.length < 10) return null; // ignore: invariant_booleans
```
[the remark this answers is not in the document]
devkabiir: @jvarness You're right, it should just be `!=10`, I probably was half asleep when made that example.

devkabiir:
@Stargator, also update the repo Readme requirements section to reflect that we require dart >=1.24.0 and support dart 2 as well.

exercises/gigasecond/test/gigasecond_test.dart
Stargator: Because we are still supporting Dart 1.0, then we need to use the polyfill (aka dart2_constant library).

exercises/armstrong-numbers/lib/example.dart
Stargator: Correct, pow returns `num` and so either we do a cast with `as` or we change the type of `sum`.

Since the values are the same, the different data types don't matter in this case. Could it be confusing to users? Yeah, probably. Would the recommendation be to change number to be of type `num`? Would that require us to change the parameter name to something else?

exercises/phone-number/lib/example.dart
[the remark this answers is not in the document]
Stargator: Addressed. I fixed this in a later commit as I cleaned up the branch.

test/exercises_test.dart
Stargator: Because contributors may find themselves switching between Dart 1.0 and Dart 2.0 `pub get` doesn't resolve the problems such a switch would usually have.

By running `pub upgrade` it forces pub to change the packages used to fit the SDK requirements.

Otherwise, contributors would have to go to **each** and *every* exercise folder and run `pub upgrade`.

An alternative to the current change, would be we create a script that just does `pub upgrade` in each exercise folder, but I'm not sure I like that.

I'm open to suggestions.

test/exercises_test.dart
```diff
@@ -30,7 +30,9 @@ Future runCmd(List<String> cmds) async {
 Future<String> getPackageName() async {
   final pubspec = new File("pubspec.yaml");
 
-  final packageName = loadYaml(await pubspec.readAsString())["name"];
```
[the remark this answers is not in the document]
devkabiir: I made the comment at wrong line, I meant to refer to line 31 and instead of var name `sourceUrl` we call it `pubspecString` since that's what it is and string is what `loadYaml` expects as its first arg.

test/exercises_test.dart
Stargator: @jvarness @devkabiir I reverted this change, just know you'd have to remove all the .dart_tool/, .pub/, .packages, and pubspec.lock files for all the exercises and the root when you switch between Dart 2 and 1.

Stargator:
@jvarness @devkabiir Ready for a re-review responded to a few of your feedback.

@devkabiir Regarding `<=3.0.0", that's what the Dart Team recommends and we are going to need to support future versions even if it means dropping support for 1.24.3 at some point.

test/exercises_test.dart
```diff
@@ -126,5 +128,5 @@ void main() {
 
       await runTest(testPath);
     }
-  }, timeout: new Timeout(new Duration(seconds: 120)));
+  }, timeout: new Timeout(new Duration(minutes: 10)));
```
devkabiir: 5 Minutes should suffice?

test/exercises_test.dart
```diff
@@ -28,7 +28,9 @@ Future runCmd(List<String> cmds) async {
 Future<String> getPackageName() async {
   final pubspec = new File("pubspec.yaml");
 
-  final packageName = loadYaml(await pubspec.readAsString())["name"];
+  final String sourceUrl = await pubspec.readAsString();
```
devkabiir: This is what I was referring to my other comment. `sourceUrl` does not hold a `URL` but `pubspec` as a `String`. `pubspecString` would be a better name.

Stargator: Got it

test/exercises_test.dart
Stargator: The more exercises we add to the longer the tests will take. So it's going to get progressively longer. I've had to extend it to 3 and half minutes. So I'd rather punt it

The point of the timeout is to stop the process in case there's an infinite loop or some problem (hardware or other software) that could case the tests to go on for too long.

Stargator:
@jvarness and @devkabiir Ready for round three. I address the variable name in exercise_test. I missed that during all my refactoring.

exercises/armstrong-numbers/lib/example.dart
jvarness: Might be a good idea to change it to `num`. I think it would be beneficial for students to understand the different data types available to them.

I think the name `number` can remain. I don't know what else you could call the parameter that would be intuitive other than `number`.

exercises/gigasecond/test/gigasecond_test.dart
jvarness: Ah, gotcha.

exercises/armstrong-numbers/lib/example.dart
Stargator: I can rename it `input` and change the type to `num`.

PR #138, opened 2018-07-29
Cleanup exercise_test runTest()

### LINKED ISSUE #137: Use `--run-skipped` flag for pub run test instead of `sed` the hacky way
I just discovered we can force run tests with `skip:true` using the `--run-skipped` flag. Seems a lot cleaner than the current approach,

devkabiir:
closes #137


devkabiir:
bad PR

PR #140, opened 2018-08-02
Provide non-empty description for Dart track.

jvarness:
Putting the finishing touches on #127 so that we're ready for the V2 launch that happened a few weeks ago 😆

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,7 +1,7 @@
 {
   "language": "Dart",
   "active": true,
-  "blurb": "",
+  "blurb": "Dart is a general purpose programming language which has seen a rise in popularity for it's uses in mobile and web application development.",
   "test_pattern": "_test\\.dart$",
   "exercises": [
     {


config.json
```diff
@@ -1,7 +1,7 @@
 {
   "language": "Dart",
   "active": true,
-  "blurb": "",
+  "blurb": "Dart is an application programming langauge developed by Google which can be used to create web applications and cross-platform mobile applications.",
```
Stargator: I really dislike describing Dart for use only as it relates to web applications and mobile apps.

I would prefer something along the lines of:

> Dart is a general purpose programming language which has seen growth in usage for web applications and cross-platform mobile applications.

jvarness: That's fair. I was trying to come up with a description that demonstrated popular use-cases for Dart. I like your description better. I'll update it when I get a chance.

Stargator: It doesn't have to be exactly like mine, free feel to make it better :)

Stargator:
@jvarness You are feel to merge this in.

PR #144, opened 2018-08-04
Add 'math' topic to relevant exercises in track config

kytrinyx:
The new site lets people filter optional exercises by topic, and this
makes it easy for people to skip past them.

This is important since the math-y exercises are not key to learning
the language, and a lot of people find them intimidating and demotivating.

This updates any 'mathematics' topics to be 'math' since it's shorter
and won't get cropped in the UI.

It also removes the math topics for any exercises that are not in the
canonical math-y list.

exercism/exercism.io#4110

### DIFF
--- a/config.json
+++ b/config.json
@@ -53,6 +53,7 @@
         "equality",
         "integers",
         "logic",
+        "math",
         "strings"
       ]
     },
@@ -64,7 +65,7 @@
       "difficulty": 2,
       "topics": [
         "integers",
-        "mathematics"
+        "math"
       ]
     },
     {
@@ -127,7 +128,6 @@
       "unlocked_by": null,
       "difficulty": 3,
       "topics": [
-        "mathematics",
         "time"
       ]
     },
@@ -140,7 +140,6 @@
       "topics": [
         "control_flow_conditionals",
         "equality",
-        "mathematics",
         "strings"
       ]
     },


jvarness:
Just out of curiousity @kytrinyx, why is math being removed from raindrops and not being added to leap?

For both exercises, you have to be able to use the modulo operator, which can be tricky to grasp and involves a good understanding of integer division and remainders.

kytrinyx:
For both `leap` and `raindrops` there's basic arithmetic involved, but the exercise itself isn't _about_ math, they've both got a story that they tell.

If you think it would be useful, then I think adding `integer_division` or just `division` or `modulo` would be fine--I just want to reserve the worth `math` for the exercises that are _about_ math rather than just relying on them.

Stargator:
@jvarness Would you want to use `integer_division`, `division` or `modulo` for those exercises as suggested?

jvarness:
I wasn't aware those tags even existed! That would be a great idea I think!

Stargator:
@jvarness As a maintainer feel free to suggest a tag/change.

kytrinyx:
Awesome. I'm going to merge this, but do feel free to make up new tags. The only tag that we've got any rules around is 'math', anything else is up to you.

Stargator:
@kytrinyx Thanks for the update, I appreciate it when issues that affect multiple repos result in related issues or PRs being opened in the respective repos.

I would have preferred whatever change @jvarness would suggest to be considered before merging this PR in. I don't think this PR was holding anything back, correct?

@jvarness is in the process of moving, so I understood his time may be limited. However, this is such a small change another PR could be opened with whatever tag(s) suggested,

Thanks again!

jvarness:
Yea sorry guys. My internet at home isn't getting hooked up until Saturday :(

This brings up a good discussion though: obviously some languages have different features than others, so different tags will apply (for example, Java would have an `objects` tag, but I would never expect an Elm exercise to have that tag).

However, many languages have the same paradigms in them, and I think it would be a good idea for the maintainers of exercism tracks to reach a consensus on consistency if we haven't already. I know that students can see how many different "types" of exercises they have completed, and I'm not sure if the tags drive that or if the specs of the exercise drive that.

@kytrinyx do the tags affect the usage of exercism from either a student or mentor perspective?

PR #148, opened 2018-08-30
convert fetch-configlet script to dart

### LINKED ISSUE #136: Convert bash script fetch-configlet to Dart
While I think this script can currently be used on all platforms, since it seems to be written well enough, this is the **Dart** track and I think it may be a good idea to have it written in Dart.

This may mean turning it into a utility class that is stored in lib/src/ and imported by create_exercise and other scripts that would need to ensure the user has configlet or the latest version of configlet.

Or the script could stay in the bin/ but just written in Dart.

devkabiir:
Closes #136


bin/fetch_configlet.dart
```diff
@@ -0,0 +1,134 @@
...
+final String osString = Platform.operatingSystem;
+final bool isWindows = osString == 'windows';
+
+void main() async {
+  final String githubApi = 'https://api.github.com/repos/exercism/configlet/releases?per_page=1';
+  String archString = '64bit';
+
+  print('Trying to detect OS and Architecture');
+  try {
+    final Process getArch = await Process.start('uname', ['-m']);
```
devkabiir: there exists a Windows equivalent command, that would remove the dependency on bash

devkabiir:
I hope the verbose nature of the script is okay. It certainly can be improved, give me your thoughts @jvarness  @Stargator

devkabiir:
I couldn't find a version of `archive` which supports both 1.24.3 and 2.0.0 of Dart.
Windows doesn't come with a built-in utility to unzip archives to it's required. 

I also couldn't figure out if decompressing can be done with stdlib `dart:io`.

On the other hand the script looks too much effort for a very simple task, we can't create a command equivalent for windows either because we still need a tool to extract archive with.

Another issue I ran into was that `runCmd` in `utils.dart` can't actually run executables from platform `path` environment variable on windows. So I've added a workaround which should be harmless on other platforms (I think).

Try running presubmit/check_formatting.dart and tell me if it works for you guys on your platforms. @Stargator  @jvarness

Stargator:
@devkabiir Sorry for the delay, reviewing this now.

bin/fetch_configlet.dart
```diff
@@ -0,0 +1,120 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:archive/archive.dart';
+
```
Stargator: Add a variable `var _operatingSystem = Platform.operatingSystem;`

Then on the first line of `main`, add `  if (_operatingSystem == 'macos') _operatingSystem = 'mac';`

That will fix an error when running on Mac. The mac zip/tar files do not have the `os` appended at the end.

However, there is a new issue that I don't know how to solve. View the output from the console below. I'll take a deeper dive later.

```
Downloading latest release from 
https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-mac-64bit.tgz
Created: "bin/"

Created: "bin/a½xxî`ðôß&©÷çW-'è¯XÂðû!rxÓKKcîãð»¾àÍn¯^¼Ràý"

;i¾§tsÐ¸s÷\s½ {îayÅo1ÊUð{û­ÃÇ\3j{äÐ"

ê;öøØàWZÃaôlÕBÿ¼Cð¯	þáõ[ðoü[ÿÊþþbh~ØýXÙO6û1³ûÑ±cÐ2¡Â"

Failed with error:
FileSystemException: Cannot open file, path = 'bin/?$¼x6~|KjpJÇ	¯Mpë«vmÈ{wºz#1gàÖ÷7 PÎ{ÏÛ¬wwó6'»{z<W@(ü¡K Eºèé©±ý®C,l/·na9Þ\7ÞE:÷¢ËÔÕQ' (OS Error: No such file or directory, errno = 2)
```

bin/fetch_configlet.dart
```diff
@@ -0,0 +1,120 @@
...
+Future<Null> main() async {
+  exitCode = await fetch_configlet().catchError((dynamic onError) {
+    print('Failed with error:\n$onError');
+    return 1;
+  });
+
+  exit(exitCode);
+}
+
+final githubApi = Uri.parse('https://api.github.com/repos/exercism/configlet/releases?per_page=1');
```
Stargator: Recommend moving this global variable above the `main` function. And changing it to a private variable.

Stargator:
Recommend updating the CONTRIBUTING doc to show how to execute the new file.

bin/fetch_configlet.dart
devkabiir: So maybe the contents of tar need different encoding to extract?. From the console output, it's creating two outputs from the archive. And when I downloaded the [Mac version](https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-mac-64bit.tgz) it only has 1 file (configlet).

Stargator: @devkabiir I don't know. But I'll try to dig deeper into this today.

devkabiir: I've updated the script to use `mac` instead of `macos`. Please try I again.

Stargator: Just tried it again:

```bash
dart bin/fetch_configlet.dart 
Created: "bin/"
Created: "bin/a½xxî`ðôß&©÷çW-'è¯XÂðû!r│ÓKK␌îãð»¾àÍ┼¯^¼Ràý"
C⎼␊▒├␊␍: "␉␋┼/ÿû÷¿ ÿ¿þ

            ÷\⎽½ πî▒≤Å⎺1ÊUðπû­ÃÇ\3┘πäÐ" &
ê;öøØàWZÃ▒ô┌ÕBÿ¼Cð¯?Æþáõ[ð⎺ü[ÿÊþþ␉␤·ØýXÙO6û1³ûÑ±␌Ð2¡Â"
F▒␋┌␊␍ ┬␋├␤ ␊⎼⎼⎺⎼:
F␋┌␊S≤⎽├␊└E│␌␊⎻├␋⎺┼: C▒┼┼⎺├ ⎺⎻␊┼ °␋┌␊, ⎻▒├␤ = '␉␋┼/?$¼│6·≠K┘⎻JÇ        ¯M⎻ë«┴└Èπ┬º≥#1±àÖ÷7 PÎπÏÛ¬┬┬ó6'»π≥<W@
                                                                                                                  (ü¡K Eºèé©±ý®C,┌/·┼▒9Þ\7ÞE:÷¢ËÔÕQ' (OS E⎼⎼⎺⎼: N⎺ ⎽┤␌␤ °␋┌␊ ⎺⎼ ␍␋⎼␊␌├⎺⎼≤, ␊⎼⎼┼⎺ = 2)
✘-1 ·/W⎺⎼┐⎽⎻▒␌␊/G␋├H┤␉/S├▒⎼±▒├⎺⎼/␍▒⎼├ [°␊├␌␤_␌⎺┼°␋±┌␊├.␍▒⎼├≠✚ 1…9⚑ 8]
```
Not the last line with the ✘, that's how my bash displays my current folder. Before it would display it like:
`✘-254 ~/Workspace/GitHub/Stargator/dart [fetch_configlet.dart|…5⚑ 8]`

So whatever happened changed how my terminal displays text.

Closing the terminal tab and opening a new one, resolve the issue. But that's still odd. I don't recommend printing out the contents of files, if that is what is happening.

devkabiir: This is definitely caused by me printing out the file names. Although they should not cause any escaping or un-escaping. This looks to me like an encoding/decoding problem when I try to un-archive the contents. This leads to having wrong/weird file names, which in turn break your terminal. I will take another look at it soon

devkabiir:
Closing because this is too out of context for me to continue work.
We currently have many breaking changes happening, and `configlet` is part of the GitHub actions now, so this is probably not required

Stargator:
@devkabiir No problem, we can revisit this later.

PR #161, opened 2019-02-04
[Draft] Travis ci update

### LINKED ISSUE #63: Make travis-ci tests faster by only testing changed files.
The following git command prints names of changed files in the latest commit.

`git diff --name-only HEAD~1`

Latest commit corresponds to the latest commit of the PR. 
This command outputs the list of files to the `stdin`
This works even for squashed merges.
Here are my proposed changes:
- Check if any `*.dart` files have been changed
    - Run `pub get`,  `pub run test` & `pub run dart_style:format -l 120 -n .` only for that exercise.
- If there are no `*.dart` files changed (could be `*.md` or `*.json`) 
    - perform `configlet lint .` if `*.json` file has been changed (only for that exercise directory)
    - do nothing and return if  `*.md` file have been changed.

I'll start working on this as soon as I get some time, until then if anyone is interested in this, create an empty PR and grab this issue.

### LINKED ISSUE #157: Dart 2.0+ Migration
Its been almost 6 months since dart 2.0 stable was release. I think we should raise the sdk constraint for the repo and all exercises to `>=2.0.0` perhaps even `>=2.1.0`. The exercises do pass against Dart 2.0 so the changes to them should be minimal.

Thoughts? @Stargator  @jvarness.

devkabiir:
Sample PR to see if it works
See #157 
Closes #63


devkabiir:
The `.travis.yaml` will need to be updated for every new exercise. This can be automated in our `create_exercise` script

This is how the new config works:
 
1. The stage "Repo" will always run the "configlet lint" job
2. The stage "Exercises" will run following for all exercises*
   - pub get
   - dartanalyzer --fatal-warnings lib test
   - dartfmt -l 120 -n --set-exit-if-changed .
   - pub run test --run-skipped

* Is skipped when
 1. There are no changes in currently being tested exercise
 2. There are changes in current exercise but only `*.md` files have been changed

What do you guys think?
Checkout build#[488](https://travis-ci.org/exercism/dart/builds/488422012)
@Stargator  @jvarness

jvarness:
And the script will exit if any part of the build in a directory fails right? Does it fail fast if, say, anagram has a test failure?

Is there a build in travis that you could link to that would demonstrate this?

.travis.yml
```diff
@@ -2,15 +2,38 @@ language: dart
...
+    - master
+
+jobs:
+  include:
+    - stage: Repo
+      name: 'configlet lint'
+      script:
+        - ./bin/fetch-configlet
+        - ./bin/configlet version
+        - ./bin/configlet lint --track-id dart .
```
Stargator: We need to run the test for create_exercise

.travis.yml
```diff
@@ -2,15 +2,38 @@ language: dart
...
+    - stage: Exercises
+      name: 'armstrong_numbers'
+      script: ./bin/travis.sh
+      env: EXERCISE_DIR="armstrong-numbers"
+      dart: stable
+      install: true
+
+stages:
+  - Repo
+  - Exercises
```
Stargator: Neither stage Repo nor Exercises set 1.24.3 or dev to be run.

Those versions of Dart should be included so we can see how handling multiple versions of Dart would work.

bin/travis.sh
```diff
@@ -0,0 +1,55 @@
...
+cp -f "lib/example.dart" "lib/$TRAVIS_JOB_NAME.dart"
+
+echo "Running pub get"
+pub get
+
+echo "Running dartanalyzer --fatal-warnings lib test"
+dartanalyzer --fatal-warnings lib test
+
+echo "Running dartfmt -l 120 -n --set-exit-if-changed ."
+dartfmt -l 120 -n --set-exit-if-changed .
```
Stargator: Recommend replacing `dartfmt` with `pub run dart_style:format`. This would allow us to use the latest version of linter than is available in the SDK.

bin/travis.sh
```diff
@@ -0,0 +1,55 @@
...
+  echo "Skipping: No changes made to $EXERCISE_DIR"
+  exit 0
+fi
+
+
+cd "exercises/$EXERCISE_DIR"
+cp -f "lib/example.dart" "lib/$TRAVIS_JOB_NAME.dart"
+
+echo "Running pub get"
+pub get
```
Stargator: Recommend considering using `pub upgrade` because when switching between Dart versions

.travis.yml
```diff
@@ -2,15 +2,38 @@ language: dart
...
+      name: 'configlet lint'
+      script:
+        - ./bin/fetch-configlet
+        - ./bin/configlet version
+        - ./bin/configlet lint --track-id dart .
+      language: bash
+
+    - stage: Exercises
+      name: 'anagram'
+      script: ./bin/travis.sh
```
Stargator: Why not use a Dart script? 😄

devkabiir:
@jvarness I do believe it should fail fast. But I'll confirm by testing and get back to you. In a real scenario Travis will run jobs for first few exercises and schedule the rest. And if the ones already running have a failure it cancels the scheduled ones and fails the build.

devkabiir:
So it seems I have a near working script.
@jvarness  @Stargator 

The latest version does following: If any of the step fails the proceeding step is never ran
 1. Always run `configlet lint`
 2. `dartfmt` is ran from the `exercism_dart` repo. For each dart version
 3. Run analyze and tests only if anything other than `*.md` is changed


From the dummy changes I made to exercises
1. acronym is skipped because there is no change
2. anagram is skipped because the change is only a `*.md` file 
3. armstrong-numbers runs analyzer and tests and fails the build
4. bob and rest of the exercises are never iterated over (fast fail)

Build [501](https://travis-ci.org/exercism/dart/builds/499810780) runs all 3 dart env in serial order so only dart stable env is ran. This is efficient in case of failures and allows other excersim repos to schedule there jobs in the meantime. In this case we always reserve 1 job at any time.

Build [502](https://travis-ci.org/exercism/dart/builds/499824795) runs all 3 dart env in parallel order so we can know if something fails in 1 env and passes in another. In this case we always reserve 1-3 jobs.

In my opinion we should go with the serialized execution since we've already cut down the time it takes to finish a build (only changed exercises tested) we don't need parallelism.

devkabiir:
Closing this in favor of #63 possibly attracting contributions.

PR #165, opened 2019-02-28
exercism/dart:  Address Dart analyzer errors due to Dart 2.0 strong typing

Stargator:
The Dart Analyzer has been giving some errors when casting some variables or values.
They did not prevent the tests or other scripts from running successfully, but I still wanted to address them to prevent confusion.

I've addressed them and formatted the files I changed.

Additionally, I added the .dart_tool directory to the ignore file.

### DIFF
--- a/bin/create_exercise.dart
+++ b/bin/create_exercise.dart
@@ -1,5 +1,6 @@
 import "dart:async";
 import "dart:io";
+
 import "package:args/args.dart";
 import "package:dart2_constant/convert.dart" as polyfill;
 import "package:path/path.dart" show dirname;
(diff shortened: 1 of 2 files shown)


devkabiir:
This would be one of the first PRs towards a Dart 2.0 migration. I recommend adding `avoid_as` as a warning in `analysis_options`. Which might surface more places that need updating + give static linting in future.

Stargator:
@devkabiir In the analysis_options.yaml file, we have `implicit-cast` and `implicit-dynamic` as `false`.

That is intended to make the typing in the repo scripts as well as the exercises stronger and more explicit.

However, in order to keep using `implicit-cast` and `implicit-dynamic` set to `false` we need to use `as` otherwise it will be very hard to work around `dynamic` return types from the Dart API.

PR #167, opened 2019-03-12
bob: Revise exercise with updates to the specification

Stargator:
The bob exercise had a .meta folder in order to force the local exercise to maintain an older state.

Since then, the test suite has been updated to reflect the changes to the test cases, but the README and the method name called from the test suite.


Stargator:
@exercism/dart Since this has been here for about 25 days, I'm going to go ahead and merge it in **4 hours**. Let me know if you'd want some time to look it over.

PR #168, opened 2019-03-12
difference-of-squares: Revise exercise with updates to the specification

Stargator:
Minor changes in response to some Dart lint rules. The imports should be sorted alphabetically and the DifferenceOfSquares type doesn't have to be used during the declaration.

The README is updated with changes from problem-specifications.


Stargator:
@exercism/dart Since this has been here for about 25 days, I'm going to go ahead and merge it in **4 hours**. Let me know if you'd want some time to look it over.

PR #170, opened 2019-03-12
Update create exercise to handle more scenarios

Stargator:
The create_exercise script needed to be updated to handle a few things:
- [x] pubspec.yaml template was out of date
- [x] test template was out of date
- [x] test template should sort the imports alphabetically
- [x] The script should use the latest version of dart_style so the formatting is leveraging the latest lint rules or updates to existing lint rules.
- [x] The Dart style guide suggests using single quotes, so the script needs to be able to use them for strings and then handle cases where an apostrophe (') is used within a string.
- [x] When detecting an integer the script should be able to select `num`, `int`, or `double` as necessary.
- [x] In some cases when a test case is generated, the expected value is an empty List. However, the script needs to know what type of object is in the List and currently can't determine what the type is.
- [X] The generator handles strings with the `\n` escape sequence, it needs to handle more escape sequences.
- [x] For matching brackets: fails to generate necessary backslashes `\` for last test
- [x] Check the version number before re-generating the exercise

### DIFF
--- a/bin/create_exercise.dart
+++ b/bin/create_exercise.dart
@@ -6,11 +6,11 @@ import 'package:args/args.dart';
 import 'package:path/path.dart' show dirname;
 
 /// Constants
-const String scriptFilename = "create-exercise";
+const String scriptFilename = 'create-exercise';
 
 final parser = new ArgParser()
-  ..addSeparator("Usage: $scriptFilename [--spec-path path] <slug>")
-  ..addOption("spec-path", help: "The location of the problem-specifications directory.", valueHelp: 'path');
+  ..addSeparator('Usage: $scriptFilename [--spec-path path] <slug>')
+  ..addOption('spec-path', help: 'The location of the problem-specifications directory.', valueHelp: 'path');
 
 /// Helpers
 List<String> words(final String str) {
@@ -18,19 +18,19 @@ List<String> words(final String str) {
 
   return str
       .toLowerCase()
-      .replaceAll(new RegExp(r"[^a-z0-9]"), " ")
-      .replaceAll(new RegExp(r" +"), " ")
+      .replaceAll(new RegExp(r'[^a-z0-9]'), ' ')
+      .replaceAll(new RegExp(r' +'), ' ')
       .trim()
-      .split(" ");
+      .split(' ');
 }
 
 String upperFirst(final String str) {
   if (str == null || str.length == 0) return '';
 
-  final chars = str.split("");
+  final chars = str.split('');
   final first = chars.first;
 
-  return first.toUpperCase() + chars.skip(1).join("");
+  return first.toUpperCase() + chars.skip(1).join('');
 }
 
 String camelCase(String str, {bool isUpperFirst = false}) {
@@ -38,40 +38,40 @@ String camelCase(String str, {bool isUpperFirst = false}) {
   final first = parts.first;
   final rest = parts.skip(1);
 
-  return (isUpperFirst ? upperFirst(first) : first) + rest.map(upperFirst).join("");
+  return (isUpperFirst ? upperFirst(first) : first) + rest.map(upperFirst).join('');
 }
 
 String pascalCase(String str) => camelCase(str, isUpperFirst: true);
 
-String snakeCase(String str) => words(str).join("_");
+String snakeCase(String str) => words(str).join('_');
 
-String kebabCase(String str) => words(str).join("-");
+String kebabCase(String str) => words(str).join('-');
 
 /// Templates
-String exampleTemplate(String name) => """
+String exampleTemplate(String name) => '''
 class ${pascalCase(name)} {
 
 }
-""";
+''';
 
-String mainTemplate(String name) => """
+String mainTemplate(String name) => '''
 class ${pascalCase(name)} {
   // Put your code here
 }
-""";
+''';
 
-String _testCasesString = """
+String _testCasesString = '''
     test('should work', () {
       // TODO
-    });""";
+    });''';
 
 /// The test template sorts the import of packages, instantiates an instance of the exercise class, and defines the main
 /// function's group of tests.
 String testTemplate(String name) {
-  final packages = <String>[name, "test"];
+  final packages = <String>[name, 'test'];
   packages.sort();
 
-  return """
+  return '''
 import 'package:${snakeCase(packages[0])}/${snakeCase(packages[0])}.dart';
 import 'package:${snakeCase(packages[1])}/${snakeCase(packages[1])}.dart';
 
@@ -82,18 +82,18 @@ void main() {
 $_testCasesString
   });
 }
-""";
+''';
 }
 
-String pubTemplate(String name) => """
+String pubTemplate(String name) => '''
 name: '${snakeCase(name)}'
 environment:
   sdk: '>=2.0.0 <3.0.0'
 dev_dependencies:
   test: '<2.0.0'
-""";
+''';
 
-String analysisOptionsTemplate() => """
+String analysisOptionsTemplate() => '''
 analyzer:
   strong-mode:
     implicit-casts: false
@@ -112,7 +112,7 @@ linter:
     - literal_only_boolean_expressions
     - no_adjacent_strings_in_list
     - valid_regexps
-""";
+''';
 
 String testCaseTemplate(String exerciseName, Map<String, Object> testCase, {bool firstTest = true, String returnType}) {
   bool skipTests = firstTest;
@@ -131,17 +131,17 @@ String testCaseTemplate(String exerciseName, Map<String, Object> testCase, {bool
       testList.add(testCaseTemplate(exerciseName, caseObj, firstTest: skipTests, returnType: returnType));
       skipTests = false;
     }
-    String tests = testList.join("\n");
+    String tests = testList.join('\n');
 
     if (description == null) {
       return tests;
     }
 
-    return """
+    return '''
       group('$description', () {
         $tests
       });
-    """;
+    ''';
   }
 
   String description = repr(testCase['description']);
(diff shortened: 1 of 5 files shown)


Stargator:
@exercism/dart I could use some help resolving an issue when generating tests for `bob`.

Specifically these two tests that have `\r` and/or `\t` in them:
* [alternate silence](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json#L161)
* [other whitespace](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json#L193)

When create_exercise.dart processes these two tests, it transforms the `\r` and '\t` into their actual whitespace counterparts instead of `\r` and `\t` respectively.

bin/create_exercise.dart
```diff
@@ -134,46 +145,170 @@ String testCaseTemplate(String name, Map<String, Object> testCase, {bool firstTe
...
+    return '$firstChar$shortenArgs$lastChar';
+  } else {
+    return null;
+  }
 }
 
 /// `repr` takes in any object and tries to coerce it to a String in such a way that it is suitable to include in code.
 /// Based on the python `repr` function, but only works for basic types: String, Iterable, Map, and primitive types
-String repr(Object x) {
+String repr(Object x, {String typeDeclaration}) {
```
Stargator: Need to add a comment here about the named param

test/create_exercise_test.dart
```diff
@@ -128,5 +129,387 @@ void main() {
...
+    - avoid_types_as_parameter_names
+    - literal_only_boolean_expressions
+    - no_adjacent_strings_in_list
+    - valid_regexps
+'''));
+      });
+
+      group('testCaseTemplate tests', () {
+//        test('testCaseTemplate test 1', () {
+//          expect(testCaseTemplate('armstrong-numbers', ), equals('''
```
Stargator: Need to complete the tests for testCaseTemplate

Stargator:
Still working on this, with the new exercises I found some things broke, so I'm looking to resolve it.

devkabiir:
@Stargator, Do you think using [mustache](https://pub.dev/packages/mustache) templates would improve the maintenance cost and make it easy to understand for the next contributor?

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea
```
devkabiir: Recommend adding `.vscode` as well.

Stargator: .vscode is already there on line 28. I'll drop this duplicate of `.idea`

Stargator:
@devkabiir Possibly. Working on this has made me curious about redesigning it from the ground up (from a `main` function perspective not the whole file).

Currently, we create things per file. So that means we are constantly reading and re-reading the input files.

I think a better path would be to read the input files and store their contents in a class representation or some other global container.

Then we could look ahead for any tests that require throwing an error, divide up tests based on the return type of the function being tested, and possibly other fun stuff.

Stargator:
@exercism/dart On the Slack channel, [@jvarness and I had a talk](https://exercism-team.slack.com/archives/CC4BCDT7X/p1570973189010900) about tracking the versioning of exercises. I was initially just going to store it in `{exercise-name}/.meta/version` for each exercise, but since Dart provides the `pubspec.yaml` file and supports a version property there, we're leaning towards doing that.

I'm going to include that change in this update.

The information is already public via this repo existing, it just currently isn't visible to users or mentors. So having it in the pubspec.yaml shouldn't cause any issue.

Still if there are any points regarding this anyone wants to talk about, this is a good place to do it.

Stargator:
@exercism/dart This is ready for review. Lots of improvements to setting the return type from the exercise's public methods/variables.

Improved handling of non-digit and non-word characters in strings.

Generally trying to reduce the amount of manually touch-ups needed for exercise generation.

Though some test suites for exercises like minesweeper and binary-search-tree when generated are vastly different from the test suites as they are now. That's fine, the exercises don't have to be the same as the output from the create_exercise script.

bin/create_exercise.dart
```diff
@@ -6,87 +6,95 @@ import 'package:args/args.dart';
 import 'package:path/path.dart' show dirname;
 
 /// Constants
```
devkabiir: Suggestion: Use `//` when using comments as code separators, otherwise `scriptFilename` gets the doc comment as `Constants`

bin/create_exercise.dart
```diff
@@ -6,87 +6,95 @@ import 'package:args/args.dart';
...
-const String scriptFilename = "create-exercise";
+const String scriptFilename = 'create-exercise';
```
devkabiir: Recommend: Rename `scriptFilename` => `scriptFileName`

bin/create_exercise.dart
```diff
@@ -6,87 +6,95 @@ import 'package:args/args.dart';
...
 
 final parser = new ArgParser()
```
devkabiir: Recommend: Remove all `new` keywords as they are optional in Dart 2.0+

bin/create_exercise.dart
Stargator: Okay, good point. I don't think I added it in this PR, but I can make the change.

bin/create_exercise.dart
Stargator: Sure

bin/create_exercise.dart
Stargator: I was thinking about doing that in this PR, but I was afraid it would be cramming too much. But I can certainly do that.

test/create_exercise_test.dart
```diff
@@ -128,5 +129,349 @@ void main() {
...
+    - literal_only_boolean_expressions
+    - no_adjacent_strings_in_list
+    - valid_regexps
+'''));
+      });
+
+      group('testCaseTemplate tests', () {
+        test('simple test for a single case', () {
+          Map<String, dynamic> testCase = <String, dynamic>{
+            "description": "Zero is an Armstrong number",
```
devkabiir: Recommend: Replace double quotes with single quotes throught out the file, where appropriate

test/create_exercise_test.dart
```diff
@@ -128,5 +129,349 @@ void main() {
...
+          <String, dynamic>{
+            "description": "year divisible by 200, not divisible by 400 in common year",
+            "property": "leapYear",
+            "input": {"year": 1800},
+            "expected": false
+          }
+        ];
+
+        Set<dynamic> actualSet = retrieveListOfExpected(cases);
+        Set<bool> expectedSet = new Set.from(<bool>[false, true]);
```
devkabiir: Recommend: Use `Set.of([false, true])`, this allows to remove the type annotations altogether.

Example `final expectedSet = Set.of([false, true]);`

devkabiir:
> @devkabiir Possibly. Working on this has made me curious about redesigning it from the ground up (from a `main` function perspective not the whole file).
> 
> Currently, we create things per file. So that means we are constantly reading and re-reading the input files.
> 
> I think a better path would be to read the input files and store their contents in a class representation or some other global container.
> 
> Then we could look ahead for any tests that require throwing an error, divide up tests based on the return type of the function being tested, and possibly other fun stuff.

Curious to know if you'd like this PR to end here or would like to continue working on it further. 

We can certainly have a discussion on how this could be re-written from ground up in a separate issue since this PR was originally to handle more test cases.

bin/create_exercise.dart
devkabiir: I thought since we are already working on it, why not make such adjustments as well. We usually don't create PRs for such refactorings and as a result they remain in the code.

bin/create_exercise.dart
```diff
@@ -105,73 +113,311 @@ linter:
...
+
+  input.forEach((String value) {
+    if (value == r'\') {
+      result.add(value + value);
+    } else {
+      result.add(value);
+    }
+  });
+
+  return StringBuffer(result.join()).toString();
```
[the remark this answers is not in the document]
Stargator: It would, I think there was an earlier case where it didn't work, but that no longer applies.

Stargator:
> - The use of `RegExp`s, I absolutely despise the use of regexs. They are not easy to parse or understand. That being said I also do not have any alternative suggestions, do you have any thoughts on this? Perhaps more doc comments to explain?

I can surely add some comments, but RegExps are very powerful and useful. I can't think of an alternative and even if I could, I wouldn't want to re-implement the work which would delay this PR significantly.

> - Recommend moving the script to `lib\src` using `git mv` to explicitly tell the git about the move and preserve commit history. The repository is certainly a dart package and based on it's purpose `lib\src` seems to me the logical place for the script. This will also allow us o use `package:` imports instead of relative imports of the script

In Dart, scripts are generally installed in the `bin/` folder. Libraries and other supporting classes/files are stored in `lib/src`. I would not want to bury our scripts. Having them in the bin matches the Dart template for projects and makes them easily accessible for contributors new and old.

> - Tye annotations for List/Set/Map elements can also be removed if the list/set/map itself is annotated. For example `final myList = <Map<String, dynamic>>[ {'one': true} , {'two': 5} ];`. In fact the type annotation can be removed altogether in that example. The dart's type system is powerful enough to deduce that `myList` has type `List<Map<String, dynamic>>`, when in doubt whether certain type annotation should be removed or not, hover over the use of that variable, if your ide/editor can tell you the correct type so can dart.

FYI, Set literals were introduced in 2.0.0, so I'd have to change the sdk constraint and I don't think that fits within this PR. We can look at it for another issue/PR

> Curious to know if you'd like this PR to end here or would like to continue working on it further.
> 
> We can certainly have a discussion on how this could be re-written from ground up in a separate issue since this PR was originally to handle more test cases.

Yeah, @devkabiir I wouldn't want to tackle that here. The primary concern with using templates or no templates, there needs to be a way to evaluate the all expected values ahead of time, but also on a test case by test case basis.

Reasoning: Some `expected`s return empty Lists, Sets, Maps, which means they don't know what types are in the collections. So the script uses Object/dynamic, but it would be useful to go back and update that based on the expected values in later test cases.

However, some test suites have the expected type changing depending on the test case.

bin/create_exercise.dart
```diff
@@ -1,92 +1,104 @@
...
 
-  return (isUpperFirst ? upperFirst(first) : first) + rest.map(upperFirst).join("");
+  return (isUpperFirst ? upperFirst(first) : first) + rest.map(upperFirst).join('');
 }
 
+/// Converts given string to pascalCase.
 String pascalCase(String str) => camelCase(str, isUpperFirst: true);
 
-String snakeCase(String str) => words(str).join("_");
+/// Converts given string to snakeCase.
```
devkabiir:
```suggestion
/// Converts given string to snake_case.
```

bin/create_exercise.dart
```diff
@@ -1,92 +1,104 @@
...
 String camelCase(String str, {bool isUpperFirst = false}) {
   final parts = words(str);
   final first = parts.first;
   final rest = parts.skip(1);
 
-  return (isUpperFirst ? upperFirst(first) : first) + rest.map(upperFirst).join("");
+  return (isUpperFirst ? upperFirst(first) : first) + rest.map(upperFirst).join('');
 }
 
+/// Converts given string to pascalCase.
```
devkabiir:
```suggestion
/// Converts given string to PascalCase.
```

bin/create_exercise.dart
```diff
@@ -1,92 +1,104 @@
...
 String pascalCase(String str) => camelCase(str, isUpperFirst: true);
 
-String snakeCase(String str) => words(str).join("_");
+/// Converts given string to snakeCase.
+String snakeCase(String str) => words(str).join('_');
 
-String kebabCase(String str) => words(str).join("-");
+/// Converts given string to kebabCase.
```
devkabiir:
```suggestion
/// Converts given string to kebab-case.
```

devkabiir:
> FYI, Set literals were introduced in 2.0.0, so I'd have to change the sdk constraint and I don't think that fits within this PR. We can look at it for another issue/PR

https://github.com/exercism/dart/blob/master/pubspec.yaml#L3 Already points to dart 2.0.0+. In any case, my point was not to use the set literals but to omit annotations as much as possible.

> In Dart, scripts are generally installed in the `bin/` folder. Libraries and other supporting classes/files are stored in `lib/src`. I would not want to bury our scripts. Having them in the bin matches the Dart template for projects and makes them easily accessible for contributors new and old.

Reason I wanted this to be in `lib/src` is that potentially in future we might re write this and perhaps write a cli tool that would do everything all of our scripts do currently.
And in that case `bin/exercism_dart.dart` would be the only script with commands to do stuff for example
1. `exercism_dart fmt` would format (including fetching dependencies such as configlet)
2. `exercism_dart create <exercise-name> --overwrite --dry-run` Would be this script

But then again that that's open for discussion in another issue.

> Yeah, @devkabiir I wouldn't want to tackle that here. The primary concern with using templates or no templates, there needs to be a way to evaluate the all expected values ahead of time, but also on a test case by test case basis.

If you are interested in taking this further, I encourage to open another issue and put all of your findings their. That way we have all the necessary info and all the necessary edge cases in one place.

All in all, thanks for your patience with my delayed reviews 😅

Stargator:
> Reason I wanted this to be in `lib/src` is that potentially in future we might re write this and perhaps write a cli tool that would do everything all of our scripts do currently.
And in that case `bin/exercism_dart.dart` would be the only script with commands to do stuff for example
1. `exercism_dart fmt` would format (including fetching dependencies such as configlet)
2. `exercism_dart create <exercise-name> --overwrite --dry-run` Would be this script

@devkabiir If that happens, the script can be moved at that point. But on the topic of merging in our scripts, while an interesting problem. Maintaining simple scripts meant for one purpose makes it easier to pickup and use.

I am not sold on the need for a comprehensive command line tool for a single track, but we can always talk about it in more depth.

PR #172, opened 2019-04-03
Adds darts exercise

amscotti:
Hi, this PR is to adds a new exercise for `darts`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/darts)

I'm still learning Dart, so feel free to point out anything that needs to be change or improved.

Thanks! :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -242,6 +242,17 @@
         "lists",
         "math"
       ]
+    },
+    {
+      "slug": "darts",
+      "uuid": "5444c90c-9129-438a-a4be-b844cf1550e5",
+      "core": false,
+      "unlocked_by": "space-age",
+      "difficulty": 3,
+      "topics": [
+        "control_flow_conditionals",
+        "math"
+      ]
     }
   ]
 }
--- a/exercises/darts/test/darts_test.dart
+++ b/exercises/darts/test/darts_test.dart
@@ -1,0 +1,61 @@
+import 'package:darts/darts.dart';
+import 'package:test/test.dart';
+
+void main() {
+  final darts = new Darts();
+
+  group('Darts', () {
+    group('Return the correct amount earned by a dart landing in a given point in the target problem.', () {
+      test("A dart lands outside the target", () {
+        final int result = darts.score(-9, 9);
+        expect(result, equals(0));
+      }, skip: false);
+
+      test("A dart lands just in the border of the target", () {
+        final int result = darts.score(0, 10);
+        expect(result, equals(1));
+      }, skip: true);
+
+      test("A dart lands in the outer circle", () {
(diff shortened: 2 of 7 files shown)


exercises/darts/test/darts_test.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:test/test.dart';
+import 'package:darts/darts.dart';
+
+void main() {
+  final darts = new Darts();
+
+  group('Darts', () {
+    group('Return the correct amount earned by a dart landing in a given point in the target problem.', () {
+      test("A dart lands outside the target", () {
+        final num result = darts.score(-9, 9);
```
Stargator: Since the score function returns an `int` I recommend dropping the use of `num` and replacing it with `int`

exercises/darts/test/darts_test.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:test/test.dart';
+import 'package:darts/darts.dart';
```
Stargator: The darts package should be imported above the test package so that it's alphabetically sorted.

exercises/darts/test/darts_test.dart
amscotti: Sounds good, and I can make this change when I get home today. :) One thing I want to point out, this file is created by `bin/create_exercise.dart` so I wasn't sure if it was ok to change the generated file or not.

Stargator: Oh yeah, create_exercise just generates a file based on a template, but we can on this track we can modify the exercise test cases however we want. We could add, subtract, or modify tests.

Also, for other exercises create_exercise may not be able to create the tests accurately (working on that). If a test is expecting an error, then the generated test case would need to be modified to use a Dart exception class. So none of the  generated files from create_exercise are set in stone.

amscotti:
@Stargator I think I got everything updated, let me know if you see anything else you would liked changed! 😃

config.json
```diff
@@ -242,6 +242,16 @@
         "lists",
         "math"
       ]
+    },
+    {
+      "slug": "darts",
+      "uuid": "5444c90c-9129-438a-a4be-b844cf1550e5",
+      "core": false,
+      "unlocked_by": null,
```
Stargator: I think having this available after completing space-age would be good.

config.json
```diff
@@ -242,6 +242,16 @@
...
+      "difficulty": 3,
+      "topics": [
+        "math"
```
Stargator: I suggest adding control-flow-conditionals as a topic.

amscotti:
@Stargator Updated the `config.json`, let me know if you see any other changes you would like. :)

PR #175, opened 2019-04-05
.gitignore: Add rule to exclude IDE files/folder

Stargator:
@exercism/dart @amscotti 

Are there any other IDE files/folders we could ignore to make it easier for you?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 bin/configlet
 bin/configlet.exe
 
-
 ### Dart ###
 # Directory created by dart tools
 /exercises/**/pubspec.lock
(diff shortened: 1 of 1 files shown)


amscotti:
@Stargator For myself, I'm using VS code with the Dart plugged and it seems to be working well. I'm not seeing any new files being created that need to be ignored.

I do have some other topics that maybe good for "General Improvements" if you would like me to add them. Some maybe just lack of documentation and lack of familiarity with the project and Dart.

jvarness:
`.vscode` would probably be a good thing to add unless we find something useful to store in it.

Stargator:
@jvarness Sure if there's anything that comes up, feel free to make a PR or discuss it in Slack if you are unsure.

amscotti:
@jvarness It would be useful to add `.vscode` but that's not created by default (which is why it wasn't popping up for me). Only gets created if you save settings on the project workspace vs the user workspace, but it would be useful. 👍

PR #177, opened 2019-04-05
Adds two-fer exercise

amscotti:
Hi, this PR is to add a new exercise for `two-fer`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/two-fer)

I'm still learning Dart, so feel free to point out anything that needs to be change or improved.

Thanks! :)

### DIFF
--- a/exercises/two-fer/.meta/canonical-data.json
+++ b/exercises/two-fer/.meta/canonical-data.json
@@ -1,0 +1,27 @@
+{
+  "exercise": "two-fer",
+  "version": "1.2.0",
+  "cases": [
+    {
+      "description": "no name given",
+      "property": "twoFer",
+      "expected": "One for you, one for me."
+    },
+    {
+      "description": "a name given",
+      "property": "twoFer",
+      "input": {
+        "name": "Alice"
+      },
+      "expected": "One for Alice, one for me."
+    },
+    {
+      "description": "another name given",
+      "property": "twoFer",
+      "input": {
+        "name": "Bob"
+      },
+      "expected": "One for Bob, one for me."
+    }
+  ]
+}
--- a/exercises/two-fer/lib/example.dart
+++ b/exercises/two-fer/lib/example.dart
@@ -1,0 +1,1 @@
+String twofer([String name = "you"]) => "One for ${name}, one for me.";
--- a/config.json
+++ b/config.json
@@ -19,6 +19,18 @@
       ]
     },
     {
+      "slug": "two-fer",
+      "uuid": "57a75d7e-370b-4937-8dd4-429db1f26ff3",
+      "core": true,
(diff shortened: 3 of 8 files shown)


exercises/two-fer/lib/example.dart
```diff
@@ -0,0 +1,5 @@
+class TwoFer {
+  String twofer(String name) {
+    return "One for ${name is String ? name : 'you'}, one for me.";
```
jvarness: This exercise could really make good use of default parameters. You might have to add a test where no parameters are passed into the function.

jvarness:
@Stargator @amscotti what do you think about making the module purely functional here with no class body?

Stargator:
@jvarness That seems like a reasonable thing to try. I think we already have at least one other exercise that does that

amscotti:
@jvarness @Stargator Updated, let me know if this was what you were thinking for changes. I kind of feel like the `null` test case should be removed If we leave this as an early exercise. What do you think?

exercises/two-fer/lib/example.dart
```diff
@@ -0,0 +1,3 @@
+String twofer([String name = "you"]) {
+  return "One for ${name is! String ? 'you' : name}, one for me.";
```
Stargator: The return statement can be replaced with the arrow syntax: ex: `String function(Parameter param) => expression;`

Stargator:
@jvarness @amscotti I would be fine with dropping the null test case as well as making it a core exercise (then the unlocked_by value would be null and the two-fer exercise moved to be after hello-world in config.yaml).

amscotti:
@Stargator Updated, let me know if I missed anything or if you see any other changes. Thanks!

exercises/two-fer/.meta/canonical-data.json
```diff
@@ -0,0 +1,22 @@
+{
+  "exercise": "two-fer",
+  "version": "1.2.0",
+  "cases": [
+    {
```
Stargator: Please add the 'no name' test case here so it aligns with the test suite

amscotti:
@Stargator Updated, let me know if that's what you wanted. I left `input` in the JSON but without anything in it.

exercises/two-fer/.meta/canonical-data.json
```diff
@@ -0,0 +1,28 @@
...
+      "description": "no name given",
+      "property": "twoFer",
+      "input": {},
```
Stargator: I think what we need to do is drop the input and that should be it.

amscotti: Removed. :)

Stargator:
@jvarness Is there anything else from your perspective?

PR #178, opened 2019-04-06
Adds reverse-string exercise

amscotti:
Hi, this PR is to add a new exercise for `reverse-string`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/reverse-string)

Thanks! :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -31,6 +31,16 @@
       ]
     },
     {
+      "slug": "reverse-string",
+      "uuid": "f195282c-c7a3-4a85-bbc1-d4e85e62d54c",
+      "core": false,
+      "unlocked_by": "two-fer",
+      "difficulty": 1,
+      "topics": [
+        "strings"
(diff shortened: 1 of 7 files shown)


config.json
```diff
@@ -18,6 +18,16 @@
         "text_formatting"
       ]
     },
+    {
+      "slug": "reverse-string",
+      "uuid": "f195282c-c7a3-4a85-bbc1-d4e85e62d54c",
+      "core": false,
+      "unlocked_by": "hello-world",
```
Stargator: How about after two-fer?

amscotti: Sounds good to me, I can make the update after `two-fer` is merged.

amscotti:
@Stargator Updated to be unlocked by `two-fer`, let me know if you see anything else that should be changed. Thanks!

PR #179, opened 2019-04-06
Adds resistor-color exercise

amscotti:
Hi, this PR is to add a new exercise for `resistor-color`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/resistor-color)

Thanks! :)

### DIFF
--- a/exercises/resistor-color/lib/example.dart
+++ b/exercises/resistor-color/lib/example.dart
@@ -1,0 +1,5 @@
+class ResistorColor {
+  final colors = ["black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white"];
+
+  int colorCode(String color) => colors.indexOf(color);
+}
--- a/config.json
+++ b/config.json
@@ -19,6 +19,17 @@
       ]
     },
     {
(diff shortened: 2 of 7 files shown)


exercises/resistor-color/lib/example.dart
```diff
@@ -0,0 +1,5 @@
+class ResistorColor {
+  final colors = ["black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white"];
```
jvarness: You might consider moving this to the package level and making it private.

Stargator: One or the other works for me. Or both. :)

Stargator: @jvarness It looks like it needs to be public in order to fit pass a test.

jvarness:
I agree with your comments @Stargator.

@amscotti I merged your other branch so this might have a conflict now. Once you've rebased one of us will merge it.

amscotti:
@jvarness All rebased, thanks! :)

PR #180, opened 2019-04-10
Adds collatz-conjecture exercise

amscotti:
Hi, this PR is to add a new exercise for `collatz-conjecture`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/collatz-conjecture)

Thanks! :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -336,6 +336,18 @@
         "loops",
         "strings"
       ]
+    },
+    {
+      "slug": "collatz-conjecture",
+      "uuid": "8cc3b7a3-6a63-4e3b-b64d-8a07d52d51f8",
+      "core": false,
+      "unlocked_by": "two-fer",
+      "difficulty": 1,
+      "topics": [
+        "exception_handling",
+        "loops",
+        "math"
+      ]
     }
   ]
 }
--- a/exercises/collatz-conjecture/lib/example.dart
+++ b/exercises/collatz-conjecture/lib/example.dart
@@ -1,0 +1,20 @@
+class CollatzConjecture {
+  int steps(int input) {
+    if (input < 1) {
+      throw new ArgumentError('Only positive numbers are allowed');
+    }
+
+    int n = input;
+    int steps = 0;
+    while (n != 1) {
+      if (n % 2 == 0) {
+        n = n ~/ 2;
+      } else {
+        n = 3 * n + 1;
+      }
+      steps += 1;
+    }
+
+    return steps;
+  }
+}
--- a/exercises/collatz-conjecture/test/collatz_conjecture_test.dart
+++ b/exercises/collatz-conjecture/test/collatz_conjecture_test.dart
@@ -1,0 +1,44 @@
+import 'package:collatz_conjecture/collatz_conjecture.dart';
+import 'package:test/test.dart';
+
+/// We are using a predicate to better match the error message from collatzConjecture.
+/// Knowing about predicates are not needed for completing this exercise,
+/// but you can find out more info at http://dartdoc.takyam.com/articles/dart-unit-tests/#matchers
+/// if you wish to learn more about predicates and matchers used in Unit Testing for Dart.
+
+void main() {
+  final collatzConjecture = new CollatzConjecture();
+  final onlyPositive = predicate(
+      (ArgumentError e) => e is ArgumentError && e.message == 'Only positive numbers are allowed',
+      'an ArgumentError with the message "Only positive numbers are allowed"');
+
+  group('CollatzConjecture', () {
+    test('zero steps for one', () {
+      final int result = collatzConjecture.steps(1);
+      expect(result, equals(0));
+    }, skip: false);
+
+    test('divide if even', () {
+      final int result = collatzConjecture.steps(16);
+      expect(result, equals(4));
+    }, skip: true);
+
+    test('even and odd steps', () {
+      final int result = collatzConjecture.steps(12);
+      expect(result, equals(9));
+    }, skip: true);
+
+    test('large number of even and odd steps', () {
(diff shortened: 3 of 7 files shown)


config.json
```diff
@@ -313,6 +313,17 @@
         "strings",
         "type_conversion"
       ]
+    },
+    {
+      "slug": "collatz-conjecture",
+      "uuid": "8cc3b7a3-6a63-4e3b-b64d-8a07d52d51f8",
+      "core": false,
+      "unlocked_by": null,
```
Stargator: I think having it unlocked by two-fer would be good.

exercises/collatz-conjecture/test/collatz_conjecture_test.dart
```diff
@@ -0,0 +1,36 @@
...
+      expect(result, equals(9));
+    }, skip: true);
+
+    test("large number of even and odd steps", () {
+      final int result = collatzConjecture.steps(1000000);
+      expect(result, equals(152));
+    }, skip: true);
+
+    test("zero is an error", () {
+      expect(() => collatzConjecture.steps(0), throwsArgumentError);
```
Stargator: Can you have the expects check that the error message that's part of the Error says "Only positive numbers are allowed". That's in the canonical-data.json.

amscotti: Will do, I'll make the update tonight!

amscotti: @Stargator Updated, let me know if this is the right way of doing this. Couldn't find a better way.

jvarness:
It would be interesting to see if this could return a `Stream` of the sequence. One of Dart's unique features is the amazing support for asynchronous computing, yet I can't recall any of our exercises supporting async processing.

However, this is also beyond the scope of this PR 😊 just an observation.

exercises/collatz-conjecture/test/collatz_conjecture_test.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:test/test.dart';
+import 'package:collatz_conjecture/collatz_conjecture.dart';
```
Stargator: The imports need to be sorted alphabetically :)

amscotti: Ah, yes. Sorry, I always forget about that. I'll see if I can get a fix out soon, if not sometime tonight.

exercises/collatz-conjecture/lib/example.dart
```diff
@@ -0,0 +1,20 @@
+class CollatzConjecture {
+  int steps(int input) {
+    if (input < 1) {
+      throw new ArgumentError("Only positive numbers are allowed");
```
Stargator: Dart style guide prefers single-quotes. where possible please replace double-quotes with single quotes.

exercises/collatz-conjecture/test/collatz_conjecture_test.dart
Stargator: @jvarness and @amscotti I'm struggling with this.

Yes, in this case the test suite has defined a predicate and the student would not need to understand it, just accept that it works.

Since collatz-conjecture is otherwise so simple and would then become the first exercise to introduce exception handling, I would advocate for a brief description of what a predicate is and why it is there as well as a link to the Dart docs on predicates.

Just so student who are curious can learn more about it.

config.json
```diff
@@ -336,6 +336,17 @@
...
       ]
+    },
+    {
+      "slug": "collatz-conjecture",
+      "uuid": "8cc3b7a3-6a63-4e3b-b64d-8a07d52d51f8",
+      "core": false,
+      "unlocked_by": "two-fer",
+      "difficulty": 1,
+      "topics": [
+        "loops",
```
Stargator: Add exception_handling to the topics.

exercises/collatz-conjecture/test/collatz_conjecture_test.dart
jvarness: +1, adding that as a comment in the tests would be helpful I think.

Stargator: Okay, good. Sorry for the delay @amscotti

amscotti: @jvarness @Stargator I have added a comment into the tests file but please feel free to let me know if something else would be better to say. More than happy to change it to make it clearer.

exercises/collatz-conjecture/test/collatz_conjecture_test.dart
```diff
@@ -0,0 +1,43 @@
...
+
+/// We are using a predicate to better match the error message from collatzConjecture.
+/// Knowing about predicates are not needed for completing this exercise,
+/// but you can find out more info at http://dartdoc.takyam.com/articles/dart-unit-tests/#matchers
+/// if you wish to learn more about predicates and matchers used in Unit Testing for Dart.
+
+void main() {
+  final collatzConjecture = new CollatzConjecture();
+  final onlyPositive =
+      predicate((ArgumentError e) => e is ArgumentError && e.message == 'Only positive numbers are allowed');
```
Stargator: Recommend giving it a custom description by add a second parameter `, 'an ArgumentError with the message "Only positive numbers are allowed"'`.

That way when the error fails it will say:

>  Expected: throws an ArgumentError with the message "Only positive numbers are allowed"

Which is much more descriptive than the default.

amscotti:
@Stargator Updated, let me know if there are any other changes you would like.

PR #182, opened 2019-04-11
Adds isbn-verifier exercise

amscotti:
Hi, this PR is to add a new exercise for `isbn-verifier`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/isbn-verifier)

Thanks! :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -324,6 +324,18 @@
         "strings",
         "type_conversion"
       ]
+    },
+    {
+      "slug": "isbn-verifier",
+      "uuid": "c52dfe2b-5634-4c0c-8ea2-3ca86e5d9a33",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "integers",
+        "loops",
+        "strings"
+      ]
     }
   ]
 }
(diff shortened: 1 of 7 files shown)


exercises/isbn-verifier/lib/example.dart
```diff
@@ -0,0 +1,19 @@
+class IsbnVerifier {
```
Stargator: I would suggest removing the class and just allowing the method to be accessible.

exercises/isbn-verifier/lib/example.dart
```diff
@@ -0,0 +1,19 @@
...
+  final RegExp formatChecker = new RegExp(r'^\d-?\d{3}-?\d{5}-?[\dX]$');
+  final RegExp cleaner = new RegExp(r'\W');
```
Stargator: If the above is taken, then these variable should be private.

config.json
```diff
@@ -313,6 +313,18 @@
         "strings",
         "type_conversion"
       ]
+    },
+    {
+      "slug": "isbn-verifier",
+      "uuid": "c52dfe2b-5634-4c0c-8ea2-3ca86e5d9a33",
+      "core": false,
```
Stargator: I think this would be a good core exercise. Otherwise, I think this PR is ready.

amscotti: @Stargator Sounds good, where did you want to have this exercise take place in this list of core exercises? Add to the end?

Stargator: It can stay where it is.

amscotti:
@Stargator Updated.

PR #186, opened 2019-04-26
Adds prime-factors exercise

amscotti:
Hi, this PR is to add a new exercise for `prime-factors`, [here is a link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/prime-factors)

Thanks! :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -251,6 +251,21 @@
       ]
     },
     {
+      "slug": "prime-factors",
+      "uuid": "89bb2ba8-86f2-4660-8739-4141eb9e6cf5",
+      "core": false,
+      "unlocked_by": "anagram",
+      "difficulty": 3,
+      "topics": [
+        "arrays",
+        "conditionals",
+        "integers",
+        "lists",
+        "loops",
+        "math"
+      ]
+    },
+    {
       "slug": "diamond",
       "uuid": "5bbd251d-b2eb-44a1-a9a6-8543afb7511c",
       "core": false,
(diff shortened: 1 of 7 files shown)


exercises/prime-factors/test/prime_factors_test.dart
```diff
@@ -0,0 +1,45 @@
+import 'package:prime_factors/prime_factors.dart';
+import 'package:test/test.dart';
+
+void main() {
+  final primeFactors = new PrimeFactors();
+
+  group('PrimeFactors', () {
+    group('returns prime factors for the given input number', () {
+      test("no factors", () {
+        final List<Object> result = primeFactors.factors(1);
```
Stargator: The return type should be `List<int>`

amscotti:
@Stargator Updated, are there any other changes you would like?

config.json
```diff
@@ -348,6 +348,21 @@
         "loops",
         "math"
       ]
+    },
+    {
+      "slug": "prime-factors",
+      "uuid": "89bb2ba8-86f2-4660-8739-4141eb9e6cf5",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 5,
```
Stargator: I would recommend a difficulty of 3

Stargator:
@amscotti I've been thinking about where to place the exercise in the order. I would recommend making it after anagram. But I don't think it has to be a core exercise.

So the needed change is just to set unlocked by `anagram`. If possible, also move it to after anagram in the list.

That should be it.

amscotti:
@Stargator This should be all set, I make it unlocked by `anagram` and moved it up the list to be next to `anagram`. Let me know if you see any other issues you want changed.

Thanks!

PR #195, opened 2019-05-08
Set Dart 2.0 as default minimum SDK

### LINKED ISSUE #157: Dart 2.0+ Migration

Stargator:
The goal of this is to reduce the time Travis CI takes to do a build.

Currently, the Dart 2.0 builds can take up to 10 minutes. I am hoping that by setting the main pubspec.lock file that it can reduce the time of those processes.

~~This should not affect the exercises, which still support Dart 1.24.3.~~

**Note:** Dropped support for 1.24.3

Fixes #157


Stargator:
I think this will have to include upgrading the exercises to 2.0, otherwise the build will fail when it tries to run the exercise tests for 1.24.3

Stargator:
After testing, there's no improvement in the build speeds unfortunately. But considering Dart 2.0 has been out for a year, I've reworked it to drop the 1.24.3 dependency across the board.

.travis.yml
```diff
@@ -1,16 +1,25 @@
...
 
 cache:
   directories:
     - $HOME/.pub-cache
+    - .dart_tool/pub
+    - .pub/
 
-script:
-  - pub run bin/check_formatting.dart
-  - pub run test
```
jvarness: I think we need to put the tests back.

jvarness: Wait... The job appears to be running tests? How is it doing that?

Stargator: [Travis CI docs](https://docs.travis-ci.com/user/languages/dart/):
> If your package depends on the test package, pub run test will be run by default.

Stargator: If I add tests back in similar to check_formatting, then it would only run it for the first version of dart listed in the config. So I would have to explicitly state which versions to run it for.

Though it would still run the tests again, so I think it's fine as it is.

jvarness: Ah, so it's built-in to the dart job. Good deal.

PR #202, opened 2019-10-05
♻refactor (config.json) Use new names for outdated topics

### LINKED ISSUE #197: Update topics for affected exercises
Exercism now has a [Track Maintenance Dashboard](https://tracks.exercism.io/dart/master/topics).

For the Dart track, there are exercises that currently use outdated topics. For example `control_flow_conditionals` would now be `conditionals` and `control_flow_loops` are now just `loops`.

The only other topic that is outdated, which we use, is `unicode`. I'm not sure if we have a suitable replacement for that.

devkabiir:
Closes #197

Note: A suitable replacement for `unicode` topic is still missing

### DIFF
--- a/config.json
+++ b/config.json
@@ -12,7 +12,7 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "control_flow_conditionals",
+        "conditionals",
         "optional_values",
         "strings",
         "text_formatting"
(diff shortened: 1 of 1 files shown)


Stargator:
@devkabiir Feel free to drop the use of `unicode` in this PR, I think it is too specific and not general enough of a topic, which is why they don't have a replacement for it.

PR #208, opened 2019-10-16
hamming: Revise exercise with updates to the specification

Stargator:
Used the updated create_exercise to review hamming and add the version to pubspec.yaml

### DIFF
--- a/exercises/hamming/lib/example.dart
+++ b/exercises/hamming/lib/example.dart
@@ -1,7 +1,11 @@
 class Hamming {
   int distance(String a, String b) {
+    if ((a.isEmpty && b.isNotEmpty) || (a.isNotEmpty && b.isEmpty)) {
+      throw ArgumentError('no strand must be empty');
+    }
+
     if (a.length != b.length) {
-      throw new ArgumentError("DNA strands must be of equal length.");
+      throw ArgumentError('left and right strands must be of equal length');
     }
 
     int diff = 0;
--- a/exercises/hamming/test/hamming_test.dart
+++ b/exercises/hamming/test/hamming_test.dart
@@ -1,84 +1,58 @@
-import 'package:test/test.dart';
 import 'package:hamming/hamming.dart';
+import 'package:test/test.dart';
+
+final hamming = Hamming();
 
 /// Test cases match one-to-one with those described
 /// in the problem specification found here:
 /// https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json
 void main() {
-  final Hamming hamming = new Hamming();
+  final mustEqual = predicate(
+      (Error e) => e is ArgumentError && e.message == 'left and right strands must be of equal length',
+      'an ArgumentError with the message "left and right strands must be of equal length"');
+  final noEmptyStrand = predicate((Error e) => e is ArgumentError && e.message == 'no strand must be empty',
+      'an ArgumentError with the message "no strand must be empty"');
 
   group('Hamming', () {
-    test("empty strands", () {
-      final int result = hamming.distance("", "");
+    test('empty strands', () {
+      final int result = hamming.distance('', '');
       expect(result, equals(0));
     }, skip: false);
 
-    test("identical strands", () {
-      final int result = hamming.distance("A", "A");
+    test('single letter identical strands', () {
+      final int result = hamming.distance('A', 'A');
       expect(result, equals(0));
     }, skip: true);
 
-    test("long identical strands", () {
-      final int result = hamming.distance("GGACTGA", "GGACTGA");
+    test('single letter different strands', () {
+      final int result = hamming.distance('G', 'T');
+      expect(result, equals(1));
+    }, skip: true);
+
+    test('long identical strands', () {
+      final int result = hamming.distance('GGACTGAAATCTG', 'GGACTGAAATCTG');
       expect(result, equals(0));
     }, skip: true);
 
-    test("complete distance in single nucleotide strands", () {
-      final int result = hamming.distance("A", "G");
-      expect(result, equals(1));
-    }, skip: true);
-
-    test("complete distance in small strands", () {
-      final int result = hamming.distance("AG", "CT");
-      expect(result, equals(2));
-    }, skip: true);
-
-    test("small distance in small strands", () {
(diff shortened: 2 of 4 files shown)


exercises/hamming/lib/example.dart
```diff
@@ -1,7 +1,9 @@
 class Hamming {
   int distance(String a, String b) {
-    if (a.length != b.length) {
-      throw new ArgumentError("DNA strands must be of equal length.");
+    if ((a.length == 0 && b.length > 0) || (a.length > 0 && b.length == 0)) {
```
devkabiir: Suggestion: Prefer `String.isEmpty` instead of `String.length==0`, Same for `String.isNotEmpty`
```suggestion
    if ((a.isEmpty && b.isNotEmpty) || (a.isNotEmpty && b.isEmpty)) {
```

exercises/hamming/lib/example.dart
```diff
@@ -1,7 +1,9 @@
...
+      throw ArgumentError('no strand must be empty');
+    } else if (a.length != b.length) {
```
devkabiir: `else if` doesn't make sense to me here. The condition `a.length != b.length` needs to be checked regardless of any prior conditions.

Suggestion: Change to independent `if` statement.
```suggestion
    } 
    
    if (a.length != b.length) {
```

exercises/hamming/test/hamming_test.dart
```diff
@@ -1,84 +1,58 @@
...
+
+final hamming = Hamming();
 
 /// Test cases match one-to-one with those described
 /// in the problem specification found here:
 /// https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json
 void main() {
-  final Hamming hamming = new Hamming();
+  final mustEqual = predicate(
+      (ArgumentError e) => e is ArgumentError && e.message == 'left and right strands must be of equal length',
```
devkabiir: I do not understand this. If the argument to predicate already has type `ArgumentError`. There is no need for `e is ArgumentError`. Since `predicate` has generic parameters, type annotation on the function parameter is required, otherwise it is inferred as `dynamic`. Since the `predicate` is being used as a generic matcher inside a `throwsA` matcher, the type annotation for function argument e should be removed. And `e is ArgumentError` should be kept.

```suggestion
      (e) => e is ArgumentError && e.message == 'left and right strands must be of equal length',
```

exercises/hamming/test/hamming_test.dart
```diff
@@ -1,84 +1,58 @@
...
+      'an ArgumentError with the message "left and right strands must be of equal length"');
+  final noEmptyStrand = predicate((ArgumentError e) => e is ArgumentError && e.message == 'no strand must be empty',
```
devkabiir: Same as `mustEqual`
```suggestion
  final noEmptyStrand = predicate((e) => e is ArgumentError && e.message == 'no strand must be empty',
```

exercises/hamming/lib/example.dart
Stargator: That's what I initially used, but I recall getting an error. So I changed it to this usage. Reverting to those methods doesn't break anything, maybe there was some other change that caused the error?

Anyway, suggestion implemented

exercises/hamming/test/hamming_test.dart
Stargator: Because of how we have analysis_options setup (no implicit casting), I had to set the parameter to Error so it's at least a generic Error. Setting it to dynamic had the analyzer throw an error. The tests run well, but I don't want the analyzer confusing users.

exercises/hamming/test/hamming_test.dart
```diff
@@ -1,84 +1,57 @@
...
+import 'package:test/test.dart';
+
+final hamming = Hamming();
 
 /// Test cases match one-to-one with those described
 /// in the problem specification found here:
 /// https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json
 void main() {
-  final Hamming hamming = new Hamming();
+  final mustEqual = predicate((ArgumentError e) => e.message == 'left and right strands must be of equal length',
```
devkabiir: Did you forget this one or is this intentional?

devkabiir: About your comment on analyzer confusing the user. Either of the following should be sufficient
```dart
predicate((Object e) => e is ArgumentError && e.message == '...', '...');
```
```dart
predicate<dynamic>((dynamic e) => e is ArgumentError && e.message == '...', '...');
```

exercises/hamming/test/hamming_test.dart
devkabiir: About your comment on analyzer confusing the user. Either of the following should be sufficient
```dart
predicate((Object e) => e is ArgumentError && e.message == '...', '...');
```
```dart
predicate<dynamic>((dynamic e) => e is ArgumentError && e.message == '...', '...');
```

devkabiir: On the note of throwing, Do you think exercise tests should be responsible for differentiating between `Error` and `Exception`. Or should the mentor be doing that?

exercises/hamming/test/hamming_test.dart
Stargator: Hmm, I thought I did that....

Stargator: Ah... I didn't commit the change. Pushing it out now

exercises/hamming/test/hamming_test.dart
Stargator: `dynamic` doesn't work, but `Object` does. Would you prefer I use that? Since Hamming is a Core exercise, I would prefer a bit more hand-holding. So I would opt for leaving `Error` in. But I'm open to be persuaded otherwise.

> On the note of throwing, Do you think exercise tests should be responsible for differentiating between Error and Exception. Or should the mentor be doing that?
That's a good question, tracks will undergo an update to V3 within the year, so how concepts are introduced are important. If a test could reasonably incorporate both, I'd love for the Test Suite or README to include a blurb about it.

Otherwise, it's up to the student and mentor.

devkabiir: Sure, as a core exercise, it makes sense to use a more specific type `Error`.

Issue #210, opened 2019-10-16
Update existing exercises to latest versions


Stargator:
With the [updated create_exercise script](https://github.com/exercism/dart/pull/170) (and a [hotfix for it](https://github.com/exercism/dart/pull/206)), we need to update our existing exercises. Some will have new test cases and/or drop existing test cases.

It is up to the track which test cases we want, if there is anything unique to the README or test cases we want specific to this track, then the respective`.meta` folder should be created and/or updated with said information. Otherwise, don't worry about the `.meta` folder.

Feel free to comment or edit this to add your name to an exercise.

Highest Priorities (core exercises):
- [x] hello-world #228 
- [x] two-fer #220
- [X] leap #209
- [x] scrabble-score* #216 #235
- [x] armstrong-numbers #225 
- [x] difference-of-squares #238
- [ ] word-count* #218 #237 
- [X] bob #239 
- [x] space-age* #217 #236 
- [X] hamming #208
- [X] gigasecond #207 #226
- [X] raindrops #214
- [X] rna-transcription #215 #226
- [x] anagram #223 
- [ ] pascals-triangle #240 
- [ ] isbn-verifier #241
- [ ] binary-search-tree #242

Side exercises
- [x] acronym #221
- [x] allergies #222
- [X] beer-song (ensure the `.meta/version` file is deleted after pubspec.yaml is updated) #211 #226
- [ ] collatz-conjecture
- [ ] darts
- [x] diamond
- [ ] isogram
- [ ] luhn
- [ ] matching-brackets
- [ ] minesweeper
- [ ] nth-prime
- [ ] pangram
- [X] phone-number #213 #226
- [ ] prime-factors
- [x] resistor-color #229
- [ ] resistor-color-duo #234
- [x] reverse-string #212
- [ ] secret-handshake
- [x] triangle

Stargator:
@exercism/dart I am leaning towards closing this issue once the last of the current PRs is merged in.

Our main focus should be on [exercism/v3](https://github.com/exercism/v3). I just want to make sure the Core exercises are updated. If anyone wants to continue to tackle updating the remaining side exercises, please leave a comment here.

The steps are:
1) Clone the repos [exercism/dart](https://github.com/exercism/dart) and [exercism/problem-specifications](https://github.com/exercism/problem-specifications)
2) Pick a side exercise and copy it's name (ie: `binary-search-tree`)
3) Run this command in the exercism/dart root directory:
[...]

PR #211, opened 2019-10-16
Beer Song: Move version into pubspec.yml

Zureka:
Per #210, the Beer Song exercise should store the version of the exercise in the `pubspec.yml` file instead of the `.meta/version` file.


Stargator:
@Zureka Thanks. Sorry about my previous feedback when you created beer-song. I thought that was the only option and after discussing with the other maintainers decided we didn't need it when we have the pubspec.yaml file.

PR #212, opened 2019-10-16
reverse-string: Update exercise and add version info

Zureka:
Re-ran the `create_exercise` script for the `reverse-string` exercise as a part of #210.

### DIFF
--- a/exercises/reverse-string/lib/example.dart
+++ b/exercises/reverse-string/lib/example.dart
@@ -1,1 +1,1 @@
-String reverse(String value) => value.split("").reversed.join('');
+String reverse(String value) => value.split('').reversed.join('');
--- a/exercises/reverse-string/test/reverse_string_test.dart
+++ b/exercises/reverse-string/test/reverse_string_test.dart
@@ -3,29 +3,34 @@ import 'package:test/test.dart';
 
 void main() {
   group('ReverseString', () {
-    test("an empty string", () {
-      final String result = reverse("");
-      expect(result, equals(""));
+    test('an empty string', () {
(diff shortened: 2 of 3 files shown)


exercises/reverse-string/lib/example.dart
```diff
@@ -1 +1,3 @@
-String reverse(String value) => value.split("").reversed.join('');
```
Stargator: This can remain as a function without a class

exercises/reverse-string/lib/reverse_string.dart
```diff
@@ -1,3 +1,3 @@
-String reverse() {
+class ReverseString {
```
Stargator: This can remain a function instead of a class.

It's part of the process. What's generated is not necessarily the standard. Some exercises are written so a class isn't required if the logic is simple enough to be contained in a function. It also exposes users to the fact that Dart does not require classes.

exercises/reverse-string/test/reverse_string_test.dart
```diff
@@ -1,31 +1,38 @@
 import 'package:reverse_string/reverse_string.dart';
 import 'package:test/test.dart';
 
+final reverseString = ReverseString();
```
Stargator: Please update the test suite without instantiating a class to fit the example.

exercises/reverse-string/lib/reverse_string.dart
Stargator: No way to do that, at this time.

Zureka:
@Stargator Is there anything else that needs to be done for this PR?

PR #219, opened 2019-10-26
Set inactive maintainers as alumni

Stargator:
@SuperPaintman hasn't contributed since I got involved in this project and @fwip contibuted since 2017.

So I suggest we set their maintainer status as alumni. I'm not exactly sure what the implications are, but overall I believe it means they moved on from the project. @iHiD correct me if I'm wrong.

### DIFF
--- a/config/maintainers.json
+++ b/config/maintainers.json
@@ -3,7 +3,7 @@
   "maintainers": [
     {
       "github_username": "SuperPaintman",
-      "alumnus": false,
+      "alumnus": true,
       "show_on_website": false,
       "name": null,
       "link_text": null,
@@ -23,7 +23,7 @@
     },
     {
       "github_username": "fwip",
-      "alumnus": false,
+      "alumnus": true,
       "show_on_website": false,
       "name": null,
       "link_text": null,


kytrinyx:
This makes sense (and it's been on my list to do this for all the tracks).

@SuperPaintman @fwip Thank you so much for the time and effort that you put into Exercism, and good luck with your current and future endeavors ❤️ You are welcome to come back any time!

Removing them from the config file should remove them automatically from the @exercism/dart team (but our bot is a bit capricious, so I'm not actually sure if that works exactly as advertised). I'll do it manually if the bot fails.

Stargator:
@kytrinyx I saw documentation that had string values, so that's why I did used them in the initial commit.

Would setting them as alumni alsy drop them from @exercism/dart, or would we have to drop them from the config?

kytrinyx:
Setting them as alumni drops them from the team.

I will try to find and amend the documentation.

kytrinyx:
Found and fixed the docs. I wonder if at some point it was implemented, but the linter currently blows up if it's a string rather than a boolean.

iHiD:
It's implemented [here](https://github.com/exercism/website/blob/master/app/services/git/sync_maintainers.rb#L63) as accepting true/false or a string. I think true/false are preferred in normal cases though.

fwip:
Thanks! I appreciate all the work that you all have done, too. <3

Stargator:
@fwip you are always welcome to come back or contribute if you want. Thanks for all your time and considerable work.

PR #223, opened 2019-11-11
anagram: Revise exercise with updates to the specification

Stargator:
Per #210, ran create_exercise to update anagram and add the version to pubspec.yaml

### DIFF
--- a/exercises/anagram/lib/example.dart
+++ b/exercises/anagram/lib/example.dart
@@ -2,7 +2,7 @@ class Anagram {
   List<String> findAnagrams(String subject, List<String> candidates) {
     Map<String, int> subjectCharsMap = countCharacters(subject);
 
-    List<String> matchTracker = new List<String>();
+    List<String> matchTracker = List<String>();
 
     for (String possible in candidates) {
       if (possible.toLowerCase() != subject.toLowerCase()) {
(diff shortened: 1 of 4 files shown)


exercises/anagram/test/anagram_test.dart
```diff
@@ -16,70 +16,80 @@ void main() {
 }
 
 void simpleTests() {
-  test("no matches", () {
-    final List<String> result = anagram.findAnagrams("diaper", ["hello", "world", "zombies", "pants"]);
+  test('no matches', () {
+    final List<String> result = anagram.findAnagrams('diaper', <String>['hello', 'world', 'zombies', 'pants']);
```
devkabiir: `<String>` type annotation on the list literal is not required. `findAnagrams` only accepts List<String> as it's second argument. Was this added by the create_exercise tool?

devkabiir: Same for rest of the list literals in this file.

exercises/anagram/lib/example.dart
```diff
@@ -2,7 +2,7 @@ class Anagram {
   List<String> findAnagrams(String subject, List<String> candidates) {
     Map<String, int> subjectCharsMap = countCharacters(subject);
 
-    List<String> matchTracker = new List<String>();
+    List<String> matchTracker = List<String>();
```
devkabiir:
```suggestion
    final List<String> matchTracker = [];
```
OR
```suggestion
    final matchTracker = <String>[];
```

PR #229, opened 2019-11-18
resistor-color: Revise exercise with updates to the specification

Stargator:
Per #210, ran create_exercise to update resistor-color and add the version to pubspec.yaml

### DIFF
--- a/exercises/resistor-color/test/resistor_color_test.dart
+++ b/exercises/resistor-color/test/resistor_color_test.dart
@@ -1,30 +1,43 @@
 import 'package:resistor_color/resistor_color.dart';
 import 'package:test/test.dart';
 
-void main() {
-  final resistorColor = new ResistorColor();
+final resistorColor = ResistorColor();
 
+void main() {
   group('ResistorColor', () {
     group('Color codes', () {
-      test("Black", () {
-        final int result = resistorColor.colorCode("black");
+      test('Black', () {
+        final int result = resistorColor.colorCode('black');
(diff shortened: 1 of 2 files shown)


exercises/resistor-color/test/resistor_color_test.dart
```diff
@@ -1,30 +1,31 @@
...
         expect(result, equals(3));
       }, skip: true);
     });
 
-    test("Colors", () {
+    test('Colors', () {
       final List<String> result = resistorColor.colors;
-      expect(result, equals(["black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white"]));
+      expect(result,
+          equals(<String>['black', 'brown', 'red', 'orange', 'yellow', 'green', 'blue', 'violet', 'grey', 'white']));
```
devkabiir: Add a trailing comma to the list to give it a better formatting.
Alternatively add a trailing comma after second argument of `expect` so both arguments can be on their own line, which might produce a better formatting
```suggestion
          equals(<String>['black', 'brown', 'red', 'orange', 'yellow', 'green', 'blue', 'violet', 'grey', 'white',]));
```
OR
```suggestion
          equals(<String>['black', 'brown', 'red', 'orange', 'yellow', 'green', 'blue', 'violet', 'grey', 'white']),);
```

Stargator: Thanks, I'll take the former suggestion :)

PR #244, opened 2020-01-18
Adds robot-simulator exercise

Corentin-Leffy:
👋 

This PR is to add a new exercise for robot-simulator, here is a [link to the problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/robot-simulator).

I'm a Dart newbie, so feel free to point out anything that needs to be change or improved.

Thanks! 😄

### DIFF
--- a/config.json
+++ b/config.json
@@ -456,6 +456,19 @@
         "bitwise_operations",
         "integers"
       ]
+    },
+    {
+      "slug": "robot-simulator",
+      "uuid": "8fffb0bb-2f71-41a9-90f2-7006020ddc19",
+      "core": false,
+      "unlocked_by": "pascals-triangle",
+      "difficulty": 4,
+      "topics": [
+        "classes",
+        "enumerations",
+        "logic",
+        "loops"
+      ]
     }
   ]
 }
--- a/exercises/robot-simulator/lib/example.dart
+++ b/exercises/robot-simulator/lib/example.dart
@@ -1,0 +1,84 @@
+import 'package:robot_simulator/orientation.dart';
+import 'package:robot_simulator/position.dart';
+
+class Robot {
+  Orientation orientation;
+  Position position;
+
+  Robot(this.position, this.orientation);
+
+  void move(String instructions) {
+    instructions.split("").forEach((character) {
+      switch (character) {
+        case "R":
+          _rotateClockwise();
+          break;
+        case "L":
+          _rotateCounterClockwise();
+          break;
+        case "A":
+          _advance();
+          break;
+        default:
+          throw StateError("unexpected instruction $character");
+      }
+    });
+  }
+
+  void _rotateClockwise() {
+    switch (orientation) {
+      case Orientation.north:
+        orientation = Orientation.east;
+        break;
+      case Orientation.east:
+        orientation = Orientation.south;
+        break;
+      case Orientation.south:
+        orientation = Orientation.west;
+        break;
+      case Orientation.west:
+        orientation = Orientation.north;
+        break;
+      default:
+        throw StateError("unexpected orientation $orientation");
+    }
+  }
+
+  void _rotateCounterClockwise() {
+    switch (orientation) {
+      case Orientation.north:
+        orientation = Orientation.west;
+        break;
+      case Orientation.west:
+        orientation = Orientation.south;
+        break;
+      case Orientation.south:
+        orientation = Orientation.east;
+        break;
+      case Orientation.east:
+        orientation = Orientation.north;
+        break;
+      default:
+        throw StateError("unexpected orientation $orientation");
+    }
+  }
+
+  void _advance() {
+    switch (orientation) {
+      case Orientation.north:
+        position.y++;
+        break;
+      case Orientation.west:
+        position.x--;
+        break;
+      case Orientation.south:
+        position.y--;
+        break;
+      case Orientation.east:
+        position.x++;
+        break;
+      default:
+        throw StateError("unexpected orientation $orientation");
+    }
+  }
+}
--- a/exercises/robot-simulator/lib/orientation.dart
+++ b/exercises/robot-simulator/lib/orientation.dart
@@ -1,0 +1,1 @@
+enum Orientation { north, east, south, west }
--- a/exercises/robot-simulator/lib/position.dart
+++ b/exercises/robot-simulator/lib/position.dart
@@ -1,0 +1,20 @@
+class Position {
+  int x, y;
+
+  Position(this.x, this.y);
+
+  @override
+  bool operator ==(Object other) => other is Position && other.x == this.x && other.y == this.y;
+
+  @override
+  int get hashCode {
+    final int prime = 31;
+    int result = 1;
+    result = prime * result + x;
+    result = prime * result + y;
+    return result;
+  }
+
+  @override
+  String toString() => "[x: $x, y: $y]";
+}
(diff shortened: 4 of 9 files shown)


exercises/robot-simulator/lib/orientation.dart
```diff
@@ -0,0 +1 @@
+enum Orientation { NORTH, EAST, SOUTH, WEST }
```
devkabiir: Recommend: Change to lowerCamelCase
```suggestion
enum Orientation { north, east, south, west }
```

In dart 1: UPPERCASE was preferred style for constants. In dart 2 its lowerCamelCase

https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names

exercises/robot-simulator/lib/example.dart
```diff
@@ -0,0 +1,68 @@
...
+
+class Robot {
+  Orientation orientation;
+  Position position;
+
+  Robot(this.position, this.orientation);
+
+  void move(String instructions) {
+    instructions.split("").forEach((character) {
+      if (character == "R") _rotateClockwise();
```
devkabiir: Suggestion: `if (character == "trigger") _action();` this is a non-returning statement.
I'm assuming your intentions are to stop checking for other "trigger"s as soon as one is found and it's corresponding action is executed.

So if `R` is the current character and `_rotateClockwise()` is the action being performed, the closure still continues to check for "L" and "A".

My suggestion is use a switch statement here as well. A switch statement makes it clear that only a single case is being handled.
An alternative would be to change to form of `if` statements to 
`if (character == "trigger") return _action();`

config.json
```diff
@@ -456,6 +456,19 @@
         "bitwise_operations",
         "integers"
       ]
+    },
+    {
+      "slug": "robot-simulator",
+      "uuid": "8fffb0bb-2f71-41a9-90f2-7006020ddc19",
+      "core": false,
+      "unlocked_by": "pascals-triangle",
+      "difficulty": 6,
```
devkabiir: Suggestion: `6` is a bit higher, i would go for `4` or `5`

config.json
```diff
@@ -456,6 +456,19 @@
         "bitwise_operations",
         "integers"
       ]
+    },
+    {
+      "slug": "robot-simulator",
+      "uuid": "8fffb0bb-2f71-41a9-90f2-7006020ddc19",
+      "core": false,
+      "unlocked_by": "pascals-triangle",
```
devkabiir: @exercism/dart, Any suggestions for this?

exercises/robot-simulator/test/robot_simulator_test.dart
```diff
@@ -0,0 +1,129 @@
+import 'package:robot_simulator/orientation.dart';
+import 'package:robot_simulator/position.dart';
```
devkabiir: I believe only a single `.dart` file is uploaded when someone submits a solution (`robot_simulator.dart`). Unless I'm mistaken, `orientation.dart` and `position.dart` should not be required and for the example.dart to pass CI, they can be made exported from `example.dart`

Or, are `orientation.dart` and `position.dart` always provided?

exercises/robot-simulator/lib/example.dart
```diff
@@ -0,0 +1,68 @@
...
+        break;
+      case Orientation.EAST:
+        orientation = Orientation.SOUTH;
+        break;
+      case Orientation.SOUTH:
+        orientation = Orientation.WEST;
+        break;
+      case Orientation.WEST:
+        orientation = Orientation.NORTH;
+        break;
```
devkabiir: Suggestion: Add a default case for all switch cases and throw a `StateError("unexpected orientation $orientation")`.

exercises/robot-simulator/lib/position.dart
```diff
@@ -0,0 +1,11 @@
+class Position {
+  num x, y;
+
+  Position(this.x, this.y);
+
+  @override
+  bool operator ==(Object other) => other is Position && other.x == this.x && other.y == this.y;
```
devkabiir: Recommend: Also override `hashCode` 
Or, If `position1 == position2` is not being used at all. Remove this override. By default dart uses `identical(position1, position2)` which suffices for example use case

exercises/robot-simulator/test/robot_simulator_test.dart
Corentin-Leffy: `orientation.dart` and `position.dart` are provided in Java and Kotlin tracks for example.

exercises/robot-simulator/lib/position.dart
Corentin-Leffy: I'm using the `equals` method in the test file when I compare both positions (and I don't want to use `identical` because it's not the same object).

exercises/robot-simulator/lib/example.dart
```diff
@@ -0,0 +1,88 @@
...
+          break;
+        case "L":
+          _rotateCounterClockwise();
+          break;
+        case "A":
+          _advance();
+          break;
+        default:
+          throw StateError("unexpected instruction $character");
+          break;
```
devkabiir: Suggestion: Remove unnecessary `break;` here.

exercises/robot-simulator/lib/example.dart
```diff
@@ -0,0 +1,88 @@
...
+        break;
+      case Orientation.south:
+        orientation = Orientation.west;
+        break;
+      case Orientation.west:
+        orientation = Orientation.north;
+        break;
+      default:
+        throw StateError("unexpected orientation $orientation");
+        break;
```
devkabiir: Suggestion: Remove unnecessary `break;` here.

exercises/robot-simulator/lib/example.dart
```diff
@@ -0,0 +1,88 @@
...
+      case Orientation.south:
+        orientation = Orientation.east;
+        break;
+      case Orientation.east:
+        orientation = Orientation.north;
+        break;
+      default:
+        throw StateError("unexpected orientation $orientation");
+        break;
```
devkabiir: Suggestion: Remove unnecessary `break;` here.

exercises/robot-simulator/lib/example.dart
```diff
@@ -0,0 +1,88 @@
...
+      case Orientation.south:
+        position.y--;
+        break;
+      case Orientation.east:
+        position.x++;
+        break;
+      default:
+        throw StateError("unexpected orientation $orientation");
+        break;
```
devkabiir: Suggestion: Remove unnecessary `break;` here.

config.json
devkabiir: Let's wait for another day and then we shall merge.

exercises/robot-simulator/test/robot_simulator_test.dart
Stargator: @Corentin-Leffy That does not seem to be true. For both the Kotlin and Java tracks, their test files implicitly inherit the source files in the `main` directory. Dart does not do that.

I would prefer we did not force users to solve an exercise in a certain way. The canonical-data.json does not specify these classes to be created, so in **my** opinion it should be left to the students to design the solution files.

devkabiir: Do you mean Kotlin/Java tests implicitly import `orientation.dart` and `position.dart` equivalents?. If that's the case they should be dropped in our tests.

This however is also a good opportunity to have `library`, `part`, `part of` directives be used in our example and have them be mentioned as a hint.

If we do not use `Orientation` and `Position` at all, then how do we write type safe tests?

Thoughts?

PR #248, opened 2020-04-14
fix spelling

wiredancer:


config.json
```diff
@@ -1,7 +1,7 @@
 {
   "language": "Dart",
   "active": true,
-  "blurb": "Dart is a general purpose programming language which has seen a rise in popularity for it's uses in mobile and web application development.",
+  "blurb": "Dart is a general purpose programming language which has seen a rise in popularity for its uses in mobile and web application development.",
```
Stargator, line 4: I would suggest also changing `uses` to `usages`. But that's my opinion.

Stargator:
@wiredancer Thanks.

Issue #257, opened 2020-08-09
gigasecond: Confusing Initial File


jameschensmith:
As seen [here](https://github.com/exercism/dart/blob/6e14fb9c07c7277caf449c95b499571f12d48078/exercises/gigasecond/lib/gigasecond.dart#L1-L3), the library template starts the user off with a class. This is confusing because in [the tests](https://github.com/exercism/dart/blob/6e14fb9c07c7277caf449c95b499571f12d48078/exercises/gigasecond/test/gigasecond_test.dart) (1) there is no instantiation of the class, (2) the tests call the function `add(DateTime)`.

This should be fixed by either (1) refactoring the tests to call the method from a class, (2) refactor the library template to start out with the function called by the tests.

Stargator:
@james-r-smith Or the user can make the changes necessary to pass the test. Exercism is not meant to teach people programming concepts, just serve as a way to introduce experienced programmers to another language.

Still, it would't hurt to make this small change.. If anyone is interested in tackling this, I see no issue with it.

Zureka:
I can look into this one. Am I correct to assume that the only required change would be to update [the starter code](https://github.com/exercism/dart/blob/6e14fb9c07c7277caf449c95b499571f12d48078/exercises/gigasecond/lib/gigasecond.dart#L1-L3) from an empty class to an empty version of the function in [the example](https://github.com/exercism/dart/blob/6e14fb9c07c7277caf449c95b499571f12d48078/exercises/gigasecond/lib/example.dart)? If that's the case I can have a PR submitted rather quickly to fix this.

devkabiir:
Fixed in #258

PR #258, opened 2020-10-02
Change from empty class to empty function

Zureka:
Updates the starter code of the `gigasecond` exercise to match the tests (and the example solution) as mention in #257


devkabiir (approved):
Thank you for your contribution 🚀

Issue #252, opened 2020-06-01
minesweeper: Delete useless constructor at the top of the test file


jmrunkle:
https://github.com/exercism/dart/blob/14858895d53b43ec91cfb1138ff82ecf29a2b352/exercises/minesweeper/test/minesweeper_test.dart#L4

All of the actual tests use a constructor that takes the input board. It makes no sense to keep this one and it was probably left in place by accident after copying and pasting a different exercise (then modifying slightly without a second thought).

Stargator:
@jmrunkle Thanks. It was accidentally left behind because our exercise generator creates a global instance. We'll track this and try to address it

sisminnmaw:
Hello @Stargator @devkabiir 
I am here because of hunting Hacktoberfest :D.
For test file, it's not bug, just like a typo errors or unused methods.

other test files use functions from object variables.
In minesweeper test file, declare object variable
`final myMineSweeper = new Minesweeper()`,
and didn't use variable.
used
`List<String> result = new Minesweeper([]).annoted;`
So, we can remove variable declaration as reporter said or we can replace new Object calling with variable name
likd
`List<String> result = myMineSweeper.annoted;`

as i just mentioned above, i wanna take this issue. i already tested.
![image](https://user-images.githubusercontent.com/5299365/95173662-ef68c400-07f3-11eb-8585-189e4416eb3f.png)


otherwise this issue still open for your exercise generator?

Stargator:
@sisminnmaw feel free to resolve this as you see fit. Let me know if you have any questions.

PR #260, opened 2020-10-07
#252 remove useless constructor

sisminnmaw:
remove useless constructor and already tested.
for issue #252 . it's not even a bug.

image


devkabiir:

PR #261, opened 2020-10-07
[diamond]: Bump exercise

Zureka:
Bumps Diamond exercise for #210

### DIFF
--- a/exercises/diamond/test/diamond_test.dart
+++ b/exercises/diamond/test/diamond_test.dart
@@ -1,87 +1,120 @@
-import 'package:test/test.dart';
 import 'package:diamond/diamond.dart';
+import 'package:test/test.dart';
+
+final diamond = Diamond();
 
 void main() {
-  final diamond = new Diamond();
-
   group('Diamond', () {
-    test("Degenerate case with a single 'A' row", () {
-      final List<String> result = diamond.rows("A");
-      expect(result, equals(["A"]));
+    test('Degenerate case with a single "A" row', () {
+      final List<String> result = diamond.rows('A');
+      expect(
+        result,
+        equals([
+          'A',
+        ]),
+      );
     }, skip: false);
 
-    test("Degenerate case with no row containing 3 distinct groups of spaces", () {
-      final List<String> result = diamond.rows("B");
-      expect(result, equals([" A ", "B B", " A "]));
-    }, skip: true);
-
-    test("Smallest non-degenerate case with odd diamond side length", () {
-      final List<String> result = diamond.rows("C");
-      expect(result, equals(["  A  ", " B B ", "C   C", " B B ", "  A  "]));
-    }, skip: true);
-
-    test("Smallest non-degenerate case with even diamond side length", () {
-      final List<String> result = diamond.rows("D");
-      expect(result, equals(["   A   ", "  B B  ", " C   C ", "D     D", " C   C ", "  B B  ", "   A   "]));
-    }, skip: true);
-
-    test("Largest possible diamond", () {
-      final List<String> result = diamond.rows("Z");
+    test('Degenerate case with no row containing 3 distinct groups of spaces', () {
+      final List<String> result = diamond.rows('B');
       expect(
-          result,
-          equals([
-            "                         A                         ",
-            "                        B B                        ",
-            "                       C   C                       ",
-            "                      D     D                      ",
-            "                     E       E                     ",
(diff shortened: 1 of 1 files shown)


exercises/diamond/test/diamond_test.dart
```diff
@@ -1,86 +1,86 @@
...
+final diamond = Diamond();
 
+void main() {
   group('Diamond', () {
-    test("Degenerate case with a single 'A' row", () {
-      final List<String> result = diamond.rows("A");
-      expect(result, equals(["A"]));
+    test('Degenerate case with a single \'A\' row', () {
+      final List<String> result = diamond.rows('A');
+      expect(result, equals(<String>['A']));
```
devkabiir, line 10: I believe adding traling commas to all instances of assertions should be more visually easy to understand. As well as adding trailing commas to the list literals as well will allow dart formatter to align th code nicely. Which would shine the diamond like shape of the result when glancing over the test file.

So adding a comma after thr `equals` call as well as th last element of the list like this
```suggestion
      expect(result, equals(<String>['A',]),);
```

Would format the code in a more visually easy to understand form
```suggestion
      expect(result, 
          equals([
              'A',
              ]),
      );
```

Also explicit type annotation `<String>` is not required.

For the very first example it doesnt seem to add much value but for later examples it definitly should.

This is of course not required but it would be beneficial for the next learner as well as the next PR reviewer.

devkabiir (changes requested):

Please let me know what you think about my comments.

exercises/diamond/test/diamond_test.dart
```diff
@@ -1,86 +1,86 @@
...
-void main() {
-  final diamond = new Diamond();
+final diamond = Diamond();
 
+void main() {
   group('Diamond', () {
-    test("Degenerate case with a single 'A' row", () {
-      final List<String> result = diamond.rows("A");
-      expect(result, equals(["A"]));
+    test('Degenerate case with a single \'A\' row', () {
```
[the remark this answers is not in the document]
Zureka: Sure thing, this is what the `bin/create_exercise.dart` script output so I figured it was the desired behavior. I can update this soon.

exercises/diamond/test/diamond_test.dart
Zureka: @devkabiir Ahh okay. The "trailing comma" formatting trick was something I picked up while learning a bit of Flutter last year but it has been awhile since I wrote any Dart code. Should a trailing comma be added after the `skip: true` as well?

devkabiir: No only for the `expect` statements (also sometimes referred to as assertions).

- Specifically 1 comma after the last element of the `<String>[]` list.

So e.g. something like `[ 'first', 'second', 'third' ]` will require a comma after the `'third'` element and will be formatted into
```dart
[
  'first',
  'second',
  'third', // here
]
```

- And 1 comma after the `equals(...)` function call.

Under the hood the dart formatter looks for a trailing comma, which is just a comma after the _last_ correct syntactical language element inside a language element container. If it finds one, it tries to _nicely_ align things on separate lines.

So this trick works for things like functions calls, function parameter definitions, collection literals, constructor calls or definitions and similarly for factory constructors and static methods. Basically anywhere you can add a comma, you can add a trailing comma. But sometimes something like `myFunc('only 1 parameter')` looks better than 
```dart
myFunc(
  'only 1 parametet',
)
````

And sometimes the other way around.

Zureka: That makes sense, especially since it's easier to see the "diamond" shape when it's formatted like that (literally the point of the exercise). I'll make that change now and rebase so it's still all in one commit.

devkabiir (approved):

Zureka:
@devkabiir  Thank you very much for taking the time to review this PR so thoroughly. Is this repository planning to participate in Hacktoberfest this year? Hacktoberfest recently updated their guidelines to help reduce the amount of spam that they were seeing in the first couple of days. 

If there aren't plans for this repository to participate this year, could you please consider adding a `hacktoberfest-accepted` label to this PR so that it counts towards my PR goal? I would really appreciate it.

devkabiir:
I only knew about the `hactoberfest` label, thanks for updating me on that.
Did that work for you?

Zureka:
It looks like that worked. Thanks again! I'm trying to do more than the minimum required this month so I'll try to submit more PRs when I have time.

PR #275, opened 2020-12-06
Update fetch-configlet with latest version from repo

Stargator:
Including latest Powershell version as well as updating the shell script version.

Issue #171, opened 2019-03-29
Create dart-analyzer


jvarness:
https://github.com/exercism/automated-mentoring-support/tree/master/docs

This is legit. We should do this for Leap to start out with.

Stargator:
I recommend integrating the dartanalyzer into this as well. That would provide some basic Dart feedback and then the analyzer would just need to be tailored to provide feedback for each exercise.

I just don't have the time to focus efforts on that.

jvarness:
That's a good idea @Stargator!

Stargator:
@exercism/dart While this is not necessary for v3 of Exercism, it would be useful for mentors like me.

AlexeyBukin:
@Stargator (sorry, I cannot mention the team), I want you to check if this issue is really `high-priority`. In that case I am ready to help!
But as a newcomer to exercism/dart I will need a _bit of_ guidance.

Stargator:
@AlexeyBukin It is a high priority, but I just won't have the time to provide guidance unless @devkabiir is available.

PR #276, opened 2020-12-08
Pin CI to latest stable Dart `2.10` & prepare for NNBD

### LINKED ISSUE #270: Update GitHub Action: Build/test against latest stable version of Dart and dev version
Given the transition to GitHub Actions, it would be a good idea to run the build and tests against the latest dev version of Dart as well as the latest stable version.

That helps us track any breaking changes that may be coming. For example, null safety is coming in 2.12 and 2.10 is currently released.

Liklainy:
* Added `matrix` with `latest` and `dev` tags
* Disabled `fail-fast` for now

Closes #270

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,8 +6,12 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        tag: [latest, dev]
+
     container:
-      image:  google/dart:latest
+      image:  google/dart:${{ matrix.tag }}
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,11 +5,14 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
-      fail-fast: false
       matrix:
-        tag: [latest, dev]
+        tag: ['2.10', latest]
+        experimental: [false]
+        include:
+          - tag: dev
+            experimental: true
 
     container:
       image:  google/dart:${{ matrix.tag }}
@@ -17,11 +20,16 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Upgrade dependencies
+      if: ${{ matrix.experimental }}
+      run: dart pub upgrade
+
     - name: Install dependencies
-      run: pub get
+      if: ${{ !matrix.experimental }}
+      run: dart pub get
 
     - name: Check Stage formatting
-      run: pub run bin/check_formatting.dart
+      run: dart run bin/check_formatting.dart
 
     - name: Run tests
-      run: pub run test
+      run: dart test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,12 +20,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Upgrade dependencies
-      if: ${{ matrix.experimental }}
-      run: dart pub upgrade
-
     - name: Install dependencies
-      if: ${{ !matrix.experimental }}
       run: dart pub get
 
     - name: Check Stage formatting
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,14 +5,10 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.experimental }}
     strategy:
+      fail-fast: false
       matrix:
-        tag: ['2.10', latest]
-        experimental: [false]
-        include:
-          - tag: dev
-            experimental: true
+        tag: ['2.10']
 
     container:
       image:  google/dart:${{ matrix.tag }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
         tag: [latest, dev]
 


.github/workflows/test.yml
```diff
@@ -6,8 +6,13 @@ jobs:
   test:
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
```
devkabiir, line 10: Please explain the reason for this flag. What are the effects of this flag and why is this necessary.

Currently the `Check Stage Formatting` is exiting with code `2` but the stage itself shows as passing `✔`. Could it be caused by this?

Does setting `fail-fast: false` deceptively show all failing stages as passing?

.github/workflows/test.yml
```diff
@@ -6,8 +6,13 @@ jobs:
...
+      matrix:
+        tag: [latest, dev]
```
devkabiir, line 12: Current stable of `2.10` is probably the last version without NNDB. I think we should have it pinned as well as `latest` &  `dev` tags. What do you think?

Ping @exercism/dart.

devkabiir (changes requested):

.github/workflows/test.yml
Liklainy: `fail-fast` description from [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast):
> When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true

Without it the `latest` job will be cancelled and not fully executed, because `dev` job failes
It can be seen in my fork [actions](https://github.com/Liklainy/dart/actions/runs/407909961) before I added `fail-fast: false`

Liklainy: Part of `Check Stage Formatting` failed with code `2` in previous [actions](https://github.com/exercism/dart/runs/1498335845) too

Exit code of `fetchConfiglet` is swallowed [inside ](https://github.com/exercism/dart/blob/master/bin/check_formatting.dart#L8) `check_formatting` and job doesn't fail
The error in fetching script probably should be fixed in #275

.github/workflows/test.yml
Liklainy: Not sure why there is a need to pin `2.10` even if it is the last version without NNDB
Pinning versions in CI will lead to inability to use new features in code in the future, there will be a need to pin next version and so on

Stargator:
@Liklainy I just updated the fetch_configlet to hopefully address the exit code issue. Please update your master branch and then issue-270 onto master. Let's see if that fixes that

.github/workflows/test.yml
Stargator: @devkabiir I'm fine either way you want to do it. As long as we would know our code works pre-NNDB, then I'm fine not pinning it.

But I also see the benefit of ensuring we are backwards compatible with pre-NNDB code (for the exercise code).

I believe NNDB is slated for version 2.12. So if we pin 2.10, then it will definitely have to be updated to 2.11 once 2.11 comes out.

Or we could wait until 2.12 is released and then go back and pin 2.11.

Any way you want to do it is good to me.

Stargator:
@Liklainy Can you update the Install dependencies step to use `pub upgrade` only when building for dev?

.github/workflows/test.yml
devkabiir: I can somewhat confidently say that the current code is compatible with `2.10`.
I don't know if `2.11` will be a real release (probably with breaking changes) or skipped over.
I know for certain `2.12` will cause issue in current code, dependencies, exercises, and solutions(example and submitted ones).

Refactoring will certainly be required for `2.12`. By convention, you don't refactor without enough tests to back you up. 
A pinned `2.10` will serve as a basic test case as well as a working state which also allows reverting to this commit via git as quickly as possible should the need arise.

We can also choose to include `// @dart=2.9` at the beginning of dart files to _opt-out_ of NNBD.  Although that's just delaying the inevitable refactoring.

NNBD will cause a lot of errors in non-migrated code for exercises, solutions, and this repo & it's not released yet. NNBD is also marked as a breaking change.  We aren't even running `dart analyze`  and don't have any preference over `analysis_options.yaml` for submitted solutions so we may have more issues in our code than we can anticipate. People solving exercises 6 months from now will most certainly be using post-NNBD dart, which might require us to include a few NNBD-specific lint rules. We also have #171, although I think it would be best to work on that post-NNBD

It's true that pinning versions prevents using new features, but it also prevents bugs to sneak-in as well as PRs that don't make changes to dart code from suddenly being broken (because the `latest` tag now points to `2.12` instead of `2.10`).

So my ideal scenario would be
 - Pin pre-NNBD stable release (currently `2.10`)
 - Pin post-NNBD stable release (not available yet)
 - Pin `latest` (once the latest tab points to `2.12`)
 - Pin `dev` (doesn't matter if this fails in CI)

I honestly think this is better suited as a separate discussion. For the purposes of this PR, I'd prefer `2.10` be pinned.
If it's possible to pin to something like `^2.10.0` then that has precedence. Since it's less likely to be containing a breaking change.

devkabiir:
Since dart `2.10` we have the new unified dart tool, so instead if using `pub upgrade` we should be using `dart pub upgrade`
See: https://medium.com/dartlang/announcing-dart-2-10-350823952bd5

Liklainy:
* Added `2.10` tag
* Changed CI to use `dart` tool
* Changed `fail-fast: false` to `continue-on-error` only for `dev` tag
So now only if `dev` job failed, workflow will continued and will be shown as [successfull](https://github.com/exercism/dart/actions/runs/421261939)
But it's not like that in every place, for example in commit it shows ❌, there is an issue actions/toolkit#399 which asks `allow-failure` implementation or fixing `continue-on-error` to behave like `allow-failure`

@Stargator, `dev` with `upgrade` is successfully [passed](https://github.com/exercism/dart/actions/runs/421223126)
I'm not sure that `upgrade` should be in CI, with it we wouldn't see the need to upgrade packages for `dev`
Removed `upgrade` in next commit to test `continue-on-error`, will add it back if it's needed

Should I squash commits?

Stargator:
@Liklainy and @devkabiir I ran `dart pub outdated --mode=null-safety and got the following result.

This was after upgrading the `test` and `yaml` packages to their null safe versions.

Screen Shot 2020-12-15 at 7 21 01 AM

Given this result and that I had to manually change the sdk version constraint to `sdk: '>=2.12.0-0 <3.0.0'` just to properly use the dev version makes me believe there's **no way** we can resolve errors with Dart dev until more packages migrate.

Additionally, I had to set the versioning for the `test` and `yaml` packages above their latest versions in order to upgrade to their null safe versions.

```yaml
  test: '>=1.15.8 <2.0.0'
  yaml: '>=2.3.0 <4.0.0'
```

So for the sake of a stable CI process, for now, I recommend dropping `dev`. That way we won't always have something failing. @devkabiir what do you think?

devkabiir:
> So for the sake of a stable CI process, for now, I recommend dropping `dev`. That way we won't always have something failing. @devkabiir what do you think?

I agree.



> For me, that means we keep the matrix setup and can even keep the 2.10 and latest tags if @devkabiir wants. So we at least have it setup to add dev when we can.

Keep the matrix setup with only `2.10` tag since the `latest` is redundant at this point and as evident by https://groups.google.com/a/dartlang.org/g/misc/c/GMHD6LOLJ1I/m/2hPXC5muAwAJ,
There won't be a Dart `2.11` so when `2.12` comes we will have to do migration. Keeping the matrix setup with only `2.10` allows for stable CI as well as prepares for next CI update.

devkabiir:
@Liklainy  I appreciate your patience and continued efforts towards this PR very much. 🚀

Liklainy:
Current changes in this PR:
* Only `2.10` tag
  * `dev` and `latest` dropped
* `matrix` setup with `fail-fast` disabled for future tags
  * `matrix` for fast adding next tags when everyhing is ready
  * `fail-fast` to allow all tags to run without cancelling in the middle when another one is failed, to see clearly when anything is working in one version and failing in another
* Using `dart` tool in CI

Stargator (approved):

devkabiir (approved):
Thank you for your contribution :rocket:.

PS: Thanks for the ping.

PR #297, opened 2021-02-17
🤖 Add file patterns to config.json

ErikSchierboom:
**⚠ This PR requires you to make a simple change before merging. ⚠**

---

To save maintainers from having to manually specify the `files` key in their exercises' `.meta/config.json` files, we are providing support for track-level patterns. See [this PR](https://github.com/exercism/docs/pull/58) for details.

This PR adds (**purposefully wrong**) file patterns to the `config.json` file. It is up to you, the track maintainers, to change these patterns to their correct value.

You can use the following placeholders:
- `%{kebab_slug}`: the `kebab-case` exercise slug (e.g. `bit-manipulation`)
- `%{snake_slug}`: the `snake_case` exercise slug (e.g. `bit_manipulation`)
- `%{camel_slug}`: the `camelCase` exercise slug (e.g. `bitManipulation`)
- `%{pascal_slug}`: the `PascalCase` exercise slug (e.g. `BitManipulation`)

We will soon update `configlet` to enable it to automatically populate the `.meta/config.json` file's `files` property, at which point we will then batch-PR updates to all tracks that have merged this PR.

## Tracking
https://github.com/exercism/v3-launch/issues/19

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,7 +11,6 @@
   "blurb": "Dart is a general purpose programming language which has seen a rise in popularity for its usages in mobile and web application development.",
   "version": 3,
   "online_editor": null,
-  "test_pattern": "_test\\.dart$",
   "files": {
     "solution": [
       "lib/%{snake_slug}.dart"
@@ -20,7 +19,7 @@
       "test/%{snake_slug}_test.dart"
     ],
     "example": [
-      "lib/example.dart"
+      ".meta/lib/example.dart"
     ],
     "exemplar": [
       "path/to/%{kebab_slug}.ext"
--- a/config.json
+++ b/config.json
@@ -22,7 +22,7 @@
       ".meta/lib/example.dart"
     ],
     "exemplar": [
-      "path/to/%{kebab_slug}.ext"
+      ".meta/exemplar.dart"
     ]
   },
   "exercises": {
--- a/config.json
+++ b/config.json
@@ -12,6 +12,20 @@
   "version": 3,
   "online_editor": null,
   "test_pattern": "_test\\.dart$",
+  "files": {
+    "solution": [
+      "path/to/%{kebab_slug}.ext"
+    ],
+    "test": [
+      "path/to/%{kebab_slug}.ext"
+    ],
+    "example": [
+      "path/to/%{kebab_slug}.ext"
+    ],
+    "exemplar": [
+      "path/to/%{kebab_slug}.ext"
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -12,6 +12,20 @@
...
+      "lib/%{snake_slug}.dart"
+    ],
+    "test": [
+      "test/%{snake_slug}_test.dart"
+    ],
+    "example": [
+      "lib/example.dart"
+    ],
+    "exemplar": [
+      "path/to/%{kebab_slug}.ext"
```
devkabiir, line 26: https://github.com/exercism/docs/pull/58 Says this is optional, so leave it as is or empty array?

config.json
```diff
@@ -12,6 +12,20 @@
   "version": 3,
   "online_editor": null,
   "test_pattern": "_test\\.dart$",
```
devkabiir, line 14: Is this still required?

devkabiir (approved):

config.json
ErikSchierboom: The field is optional because you don't _have_ to set it, but I'd set it to:

```suggestion
      ".meta/exemplar.dart"
```

config.json
```diff
@@ -12,6 +12,20 @@
...
+  "files": {
+    "solution": [
+      "lib/%{snake_slug}.dart"
```
devkabiir, line 17: https://github.com/exercism/docs/pull/58#issuecomment-780789579 mad me concerned, what happens to other support/lib/module files, someone writes as part of the solution?

config.json
devkabiir: I'm fine with your suggestion.
I have a question though, the `exemplar.dart` doesn't exist for any practice exercise, does that cause any breaking effects?

config.json
ErikSchierboom: Nothing. People can submit them in exactly the same way that they can now.

ErikSchierboom:
I've created a script that will automatically populate the `files` property based on the pattern specified in the `config.json` (which was added in this PR). If you'd like, I could run the script for your track, which would result in a PR that looks like this: https://github.com/exercism/csharp/pull/1507

In the future, we'll update `configlet` to include this functionality. For now, I can thus do this for you by running a script. Let me know if you'd like me to run the script for your track.

devkabiir:
Please do. :rocket:

PR #298, opened 2021-02-18
[v3] Move Practice Exercise example files to .meta directory

ErikSchierboom:
The example implementation files for Practice Exercises should be located in the `.meta` directory. Besides being consistent with Concept Exercises, it also has the added benefit that no extra configuration is needed to prevent the example file from being downloaded to the student (as files in the `.meta` directory are never downloaded).

See [the spec](https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md#file-example-implementation).

This PR attempts to detect these example files and move them to the exercise's `.meta` directory.

Note that this change will likely break your CI, so feel free to either merge this and fix the build later, or push a fix to this PR.  

## Tracking

https://github.com/exercism/v3-launch/issues/18


devkabiir:
This is a breaking change for `Test` workflow. The way we validate example implementations as well as tests is by changing the name of `lib/example.dart` to `lib/${snake_case}.dart` and running usual tests with dart's `pub test` or `dart test` commands in that exercise's directory. This will break that script, tracking #299

devkabiir (approved):
This is a breaking change for `Test` workflow. The way we validate example implementations as well as tests is by changing the name of `lib/example.dart` to `lib/${snake_case}.dart` and running usual tests with dart's `pub test` or `dart test` commands in that exercise's directory. This will break that script, tracking #299 


Do we really need to specify example file path for each exercise in `exercise/.meta/config.json` ?
#297 should be enough for track-level file patterns no?

ErikSchierboom:
>This is a breaking change for Test workflow. The way we validate example implementations as well as tests is by changing the name of lib/example.dart to lib/${snake_case}.dart and running usual tests with dart's pub test or dart test commands in that exercise's directory. This will break that script, tracking #299

That is unfortunate, but v3 does require a set of breaking changes, this being one of them. As tracks are now disconnected from the v2 website, it is okay to have tracks temporarily in a broken state.

> Do we really need to specify example file path for each exercise in exercise/.meta/config.json ?
#297 should be enough for track-level file patterns no?

We do. The website won't read the track-level file patterns and just look at the exercise's .meta/config.json file. This allows exercises to deviate from the default as well as makes it easier for tooling (like the test runner) to figure out what the actual files are. We will probably tooling (configlet) to automatically populate the fields.

PR #309, opened 2021-03-21
Bump dart style

Stargator:
dart_style was finally updated to depend on 2.12 of Dart. This PR updates our dependency and drops dart_style from the concept exercises since dart_style is only needed in our root pubspec.yaml for linting and CI


Stargator:
Thanks @devkabiir. I'm going to continue working on the strings concepts exercise.

PR #332, opened 2021-06-30
Add tags

ErikSchierboom:


Stargator:
@ErikSchierboom What's the process if we want to update the tags moving forward? Can we modify them directly or is there a settings file we'd have to update?

ErikSchierboom:
@Stargator With "update the tags" you mean adding to the list of available tags?

Stargator:
@ErikSchierboom I mean adding or removing. I ask because I know for some lists (like the labels for issues/PRs) they are stored in files in the repo. Not sure about these tags.

ErikSchierboom:
@Stargator The tags are a fixed list which is maintained by the Exercism staff. We don't allow adding/updating by maintainers as new tags would also require changes in several other places, including configlet and the website. We also don't want to be exhaustive with the tags, as our goal is solely to make it easier for a student to find a track given some popular criteria. If you'd like to suggest more tags, please open an issue on https://github.com/exercism/exercism/issues.

iHiD:
(You can change the tags used on your track within the predefined set just by updating this file at any time though)

PR #344, opened 2021-08-23
Set Stargator as inactive maintainer

Stargator:
As previously posted in another issue. I just don't feel like I know what's going on with Exercism or this track.

Part of it is definitely due to the rhythm of my contributions and how some initial plans for v3 have changed since I had started work.

Another part is just the frustration due to my understanding of tracks getting help for developing concepts, but that maintainers would still have a say in how the concept exercises are planned.

I don't support scenario concept exercises. I prefer straight to the point exercises that just show off the features surrounding a concept without the fluff.

And since my understanding of the track and Exercism for v3 has drifted from where I thought it would be, I can't contribute when I don't follow much of what is going on.

### DIFF
--- a/config/maintainers.json
+++ b/config/maintainers.json
@@ -13,7 +13,7 @@
     },
     {
       "github_username": "Stargator",
-      "alumnus": false,
+      "alumnus": true,
       "show_on_website": true,
       "name": null,
       "link_text": null,


iHiD:
Hi @Stargator, Thanks for your openness. Might it be worth waiting until v3 launches a week from now at which point you could try running through a track that's got Concept Exercises and decide then if things make more sense and you understand (and maybe like) how things are working? Also, as this track has no Concept Exercises, yet there will be still lots to do in evolving the track without this (e.g. test runners, representers, etc), there might be other areas where you'd like to still contribute other than these Concept Exercises?

With regards to your specific points:
- As I discussed in the previous issue, I don't feel like the idea of Concept Exercises have changed as dramatically as you feel. The very first ones were story-based - the only change was that the slug was based on the concept, not the story, which didn't work when there were multiple concepts taught by one exercise, so we changed this.
- With regards to support in making exercises, there are now dozens of Concept Exercises ([e.g. Elixir](https://github.com/exercism/elixir/tree/main/exercises/concept)), which can be forked, and [dozens of stories](https://github.com/exercism/docs/tree/main/building/tracks/stories) pre-written to use as themes. So there is a lot of support/help there both in terms of content and in terms of support from the people who have written the exercises if want to get an initial set of exercises added.

Of course, if you just want to resign as a maintainer I respect that, but it feels like a maybe unnecessary loss to Exercism for you to leave.

iHiD:
@Stargator v3 is now launched if you want to see if you still want to be part of things or not? We'd love to have you around still 🙂

Stargator:
@iHiD I can wait. I am generally too burnt out to put much energy into it right now. I can go ahead and close this until I get a fresh look at how other tracks do concept exercises.

As for your point about looking at other aspects of the track, yeah that's where I would prefer to spend my time, but as I have been the only other active maintainer aside from `devkabiir` I felt some pressure to try to get the track moving on the concept exercises which seemed like the bigger lift to me.

iHiD:
Great :)

It's fine for the track to not have concept exercises - it's still super valuable and we can incrementally improve it without. And then if you or others ever get a larger chunk of time to work on the concepts, then great.

And take care of yourself - burnout isn't fun - I know from experience. Glad to have you around still :)

iHiD:
@Stargator I tried to find the original issue about Concept Exercises that you wrote to post this in but couldn't so am just posting it here. I thought this was a nice explanation about "stories" in Concept Exercises from @AngelikaTyborska, which sort of outlines it really well (Angelika has written dozens of concept exercises for Elixir), and might possibly put your mind at ease a little more about how they're positioned.

> There is a story to each learning exercise, but the story is not that important. It's there so that you don't have to use nonsensical values "foo bar" and "1 2 3" in every exercise, but it doesn't play a big role in the exercise like it does in practice exercises.

PR #347, opened 2021-09-02
Update leap_test.dart to avoid skipping the tests

faizmokhtar:
I was working on the leap exercise locally and set the function to return to `false` by default. When, I run the unit test it passes successfully which is unexpected. Upon checking the test cases, I noticed most of the test cases are set to be skippable.

This PR should fix the issue by setting it back to `false`

### DIFF
--- a/exercises/practice/leap/test/leap_test.dart
+++ b/exercises/practice/leap/test/leap_test.dart
@@ -13,41 +13,41 @@ void main() {
     test('year divisible by 2, not divisible by 4 in common year', () {
       final bool result = leap.leapYear(1970);
       expect(result, equals(false));
-    }, skip: true);
+    }, skip: false);
 
     test('year divisible by 4, not divisible by 100 in leap year', () {
       final bool result = leap.leapYear(1996);
       expect(result, equals(true));
-    }, skip: true);
+    }, skip: false);
 
     test('year divisible by 4 and 5 is still a leap year', () {
       final bool result = leap.leapYear(1960);
       expect(result, equals(true));
-    }, skip: true);
+    }, skip: false);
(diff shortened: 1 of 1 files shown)


angelikatyborska:
Hi @faizmokhtar, thank you for your input! I am not part of the Dart maintaining team, but from another track, but I think I can help out here.

It is **desired** that in every single **practice** exercise all tests but the first one are skipped. If you take a look at any other practice exercise in this repository, you'll notice that's true.

This is because Exercism wants to encourage test-driven development when solving practice exercises.

The instructions on how to run the tests _should_ explain that. But I see that they don't.

This is the relevant file: https://github.com/exercism/dart/blob/main/exercises/shared/.docs/tests.md

You could help out by opening a PR that updates that file to mention test skips. You can use https://github.com/exercism/dart/blob/main/docs/TESTS.md as inspiration or just copy-paste the part about understanding test results.

Some context behind those documents:
- `shared/.docs/tests.md` shows up in `README.md` when you download an exercise via the CLI or on the website in the web editor in the instruction, so it needs to be relevant for both use cases.
- `docs/TESTS.md` is rendered here https://exercism.org/docs/tracks/dart/tests

faizmokhtar:
@angelikatyborska I see. That actually make sense since I noticed it was skipped on other practice exercises' tests file too.

Thanks for your reply and recommendation. I'll do that on different PR 🙂

Stargator:
Thanks @faizmokhtar for bring this to our attention, we will welcome your PR.

Thank you @angelikatyborska for doing a wonderful job explaining our approach. I thought we had the explanation in the docs at some point.

PR #348, opened 2021-09-03
Update TESTS.md (dart test should also be a command people should know about)

MSoup:
I noticed that many people have had problems using pub. I added a little excerpt about how `dart test` also works, and is possibly more beginner friendly.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install dependencies
-      run: dart pub get
+      run: dart dart pub get
 
     - name: Check Stage formatting
       run: dart run bin/check_formatting.dart
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install dependencies
-      run: dart dart pub get
+      run: dart pub get
 
     - name: Check Stage formatting
       run: dart run bin/check_formatting.dart
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -12,7 +12,7 @@ The test run summary will list the number of passed, skipped and failed tests.
 For example :
 
 ```
-❯ dart run test
+❯ dart test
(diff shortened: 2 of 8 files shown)


docs/TESTS.md
```diff
@@ -29,6 +29,19 @@ It is advised to enable tests one by one as you implement the functionality bein
...
+NOTE: Some users may run into the error:
+
+```
+command not found: pub
+```
+
+If your terminal does not know where pub is, you may need to add it to your PATH. Alternatively, try the following command from the root directory of the exercise:
+
+```
+dart test
```
Stargator, line 41: I suggest updating all the existing commands to use the newer structure. Like replace `pub run test` with `dart test`.

Stargator (changes requested):

MSoup (commented):
I think this should do it. There is a HELP.md that comes with each exercise that isn't available in here. Do you happen to know where I can access it?

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Install dependencies
-      run: dart pub get
+      run: dart dart pub get
```
Stargator: No need to duplicate the `dart` command. That won't work in GitHub Actions.

I think you just need a copy/paste, so just need to correct this.

Stargator (changes requested):
I correction and I think this would be good.

.github/workflows/test.yml
MSoup: My apologies, I caught that mistake and did a second commit that reflects that change.

Stargator (changes requested):

exercises/shared/.docs/tests.md
```diff
@@ -3,5 +3,5 @@
 To run the tests:
 
 ```sh
-$ pub run test
+$ dart run test
```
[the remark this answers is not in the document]
MSoup: Apologies for the delay. Does my latest commit seem okay?

Stargator (approved):

Stargator:
@MSoup Sorry for the delay, I am running the Actions workflow against this and if good, I'll merge it

PR #352, opened 2021-09-10
Disable test runner

ErikSchierboom:
We've had some issues getting the test runner to consistently finish running within the allotted time. Until those issues are fixed, let's disable the test runner.


iHiD (approved):

ErikSchierboom:
@Stargator Let's try to continue working on the performance in the background. It would be great if we can re-enable the test runner.

PR #353, opened 2021-09-22
set the indent size to 2 according for style guide

ymadd:


Stargator:
@ymadd Why should we adopt the styling guide's indention? What benefit would that provide?

ymadd:
@Stargator 
It seems that the online editor is not available at the moment, but if the online editor is enabled and more and more answers are given using it, there will be a difference in the environment such as indenting between local and online, and if such a difference is created, the mentors will have to read the difference in the code. In order to eliminate these differences as much as possible, I think it would be helpful to define the indentation style used in the dart format, etc., to minimize the differences and reduce the burden on the mentor.

Stargator:
@ymadd Oh! I see. You are talking about the indentation within the editor not for the exercises and practice solutions in this repo. I'll update the header for that.

I don't think this repo handles the editor, so I'm pinging @ErikSchierboom and @iHiD

PR #360, opened 2021-10-27
added github actions to check broken links

### LINKED ISSUE #357: Add or Update existing GitHub Action to check for bad links
There have been a few pull requests over the years that have identified bad links in our docs. I think it's time for us to have a workflow action that will check for any bad links so we catch them sooner.

The related PR should:
* Check for any bad links in any markdown files (the root docs directory as well as the .docs/ folder for exercises)
* Check for English (US) spelling errors

lakshya8066:
Fixes #357


Stargator:
Marking this as Hacktoberfest accepted since this is a variable update to our workflow and a great amount of research and testing that goes into it.

lakshya8066:
Thank you @Stargator :)

.github/workflows/links.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.0.9
+        with:
+          args: --verbose --no-progress **/*.md **/*.html
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        
+      - name: Create Issue From File
```
Stargator, line 22: I don't think we need to create an issue as if the Workflow fails, we'd just have the PR branch updated.

Stargator (changes requested):
Almost ready, just drop the create issue logic and I think this is good.

.github/workflows/links.yml
lakshya8066: Oh okay, I will update the PR

lakshya8066: Just to be sure, we don't need the environment variable too right?

Stargator: I don't think so.

Stargator:
@lakshya8066 It seems the Links workflow was not executed. Maybe we do need the secret token?

lakshya8066:
Right, let's add it and see...

.github/workflows/links.yml
```diff
@@ -0,0 +1,20 @@
+name: Links
+
+on:
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "00 18 * * *"
```
Stargator, line 7: @lakshya8066 I see the issue now, it's not running because There hasn't been a repository_dispatch or a workflow_dispatch event. And the cron is waiting on its schedule.

If we want it to happen during pull_requests, add `pull_request:` and I would even add `push:`

Stargator (changes requested):

.github/workflows/links.yml
```diff
@@ -0,0 +1,20 @@
+name: Links
+
+on:
+  repository_dispatch:
+  workflow_dispatch:
```
Stargator, line 5:
```suggestion
on:
  push:
  pull_request:
  repository_dispatch:
  workflow_dispatch:
```

.github/workflows/links.yml
lakshya8066: Oh yeah, that might be the issue...

Stargator:
@lakshya8066 Yep, now the action is running.

Stargator:
@lakshya8066 It's working, but there is an error for one link and yet the Action seems to be passed as Successful in the PR. I see it is returning exit code 2, is there someway to ensure it fails on error?

lakshya8066:
@Stargator yes and I believe it stops checking once it gets a broken link or maybe it is just a coincidence. 
It was originally supposed to create an issue for the broken link but even without that, tests should fail for non-zero exit codes.
I'm not sure how to achieve this, however, I think [this](https://docs.github.com/en/actions/learn-github-actions/contexts#job-status-check-functions) can be helpful. If there is a way to grab the exit code information, we edit the yml to compare the exit codes with 0 and fail the test accordingly. What do you think?

.github/workflows/links.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.0.9
+        with:
+          args: --verbose --no-progress **/*.md **/*.html
```
Stargator, line 20:
```suggestion
          args: --verbose --no-progress --require-https  **/*.md **/*.html
```

I think we should definitely check for https. I know this will cause more errors, but it'll help us catch typos in the future.

.github/workflows/links.yml
```diff
@@ -0,0 +1,22 @@
...
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
Stargator, line 22:
```suggestion
      - name: Link Checker
        uses: lycheeverse/lychee-action@v1.0.9
        with:
          args: --verbose --no-progress **/*.md **/*.html
        env:
          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
      -if: failure()
      - if: success()
```
Let's try this, [I also opened a PR on the repo for the link-checker action](https://github.com/lycheeverse/lychee-action/issues/58).

Stargator: Alternatively, the maintainer of lychee-action suggests:

```suggestion
      - name: Link Checker
        uses: lycheeverse/lychee-action@v1.0.9
        with:
          args: --verbose --no-progress **/*.md **/*.html
        env:
          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
     - name: Fail if there were link errors
        run: exit ${{ steps.lychee.outputs.exit_code }}
```

lakshya8066: I think we should check this once. The use case [here](https://github.com/lycheeverse/lychee-action/issues/20) is similar to ours

.github/workflows/links.yml
lakshya8066: Yes it will cause more errors and all websites with http will have to be changed to https to pass the test right?

Stargator: Yes, but that's acceptable. We want to verify that the Action will fail and then we apply fixes to it.

.github/workflows/links.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.0.9
+        with:
+          args: --verbose --no-progress --require-https **/*.md **/*.html
```
Stargator, line 20:
```suggestion
          args: --verbose --no-progress **/*.md **/*.html
```

The argument is not recognized for some reason, so let's drop it.

.github/workflows/links.yml
```diff
@@ -0,0 +1,24 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Link Checker
+        uses: lycheeverse/lychee-action@v1.0.9
+        with:
+          args: --verbose --no-progress **/*.md **/*.html
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      - name: Fail if there were link errors
+        run: exit ${{ steps.lychee.outputs.exit_code }}
```
Stargator, line 24:
```suggestion
      - name: Link Checker
        id: checker
        uses: lycheeverse/lychee-action@v1.0.9
        with:
          args: --verbose --no-progress **/*.md **/*.html
        env:
          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
      - name: Fail if there were link errors
        run: exit ${{ steps.checker.outputs.exit_code }}
```

No step exists by the name lychee because the step was not given an id

Stargator:
@lakshya8066 I made a few updates to resolve some errors. But now it's failing! YAY! 💯 

Last steps now:

1) Could you replace all `http` with `https`?
2) Update http://exercism.io/languages/dart/help @ErikSchierboom @iHiD, did the help pages get moved? Where should we point users?

ErikSchierboom:
@Stargator They can be accessed at https://exercism.org/docs/tracks/dart

.github/workflows/links.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Link Checker
+        id: checker
+        uses: lycheeverse/lychee-action@v1.0.9
```
Stargator, line 19: I suggest updating to the latest, a few new versions came out. The latest at this time is v1.1.1

Stargator (approved):

Stargator:
@lakshya8066 I am approving this PR and merging it. **But**, I need you to understand that normally I would require all the checks to pass. That is why I expected the URLs to be fixed in this branch so the new link checker can verify everything is good.

So do understand in the future for this repo or any future repo you contribute to. If there are fixes related to an existing PR it's probably best to do it all in one branch.

Thanks again for **all your work** and **patience**, I really appreciate finally having this action in our workflow. 😄 🥇

lakshya8066:
@Stargator absolutely, I will keep that in mind. I realized the same after creating the other PR. 
And thanks for supporting and guiding me through the PR! This was indeed an amazing learning experience. ✨
Looking forward to contributing more to the project :)

PR #365, opened 2021-11-17
Update config.json

andrewfgarrison:
Updated the key features

### DIFF
--- a/config.json
+++ b/config.json
@@ -618,7 +618,38 @@
       "name": "Type Conversion"
     }
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      icon: "web",
+      title: "Popular for Web and Mobile Development",
+      content: "Dart has the Flutter library, which can be used to quickly build web and smartphone applications."
+    },
+    {
+      icon: "cross-platform",
+      title: "Instant iOS and Android Development",
+      content: "Dart/Flutter can be converted into iOS's Swift language and Android's Kotlin language, making mobile development fast and efficient."
+    },
+    {
+      icon: "community",
+      title: "Full Documentation",
+      contents: "Dart's documentation and learning materials are constantly updated by its developers, making it easy to look up information on the language."
+    },
+    {
+      icon: "community",
+      title: "Countless Packages",
+      contents: "Dart allows users to build and publish their own packages for public use, which can be accessed from Dart directly."
+    },
+    {
+      icon: "functional",
+      title: "Highly-organized Code",
+      contents: "Dart's coding style makes it easy to read and understand."
+    },
+    {
+      icon: "fun",
+      title: "Easy to Learn",
+      contents: "Dart is easy to understand and great for beginners to programming."
+    }
+  ],
   "tags": [
     "paradigm/declarative",
     "paradigm/functional",
--- a/config.json
+++ b/config.json
@@ -632,22 +632,22 @@
     {
       "icon": "community",
       "title": "Full Documentation",
-      "contents": "Dart's documentation and learning materials are constantly updated by its developers, making it easy to look up information on the language."
+      "content": "Dart's documentation and learning materials are constantly updated by its developers, making it easy to look up information on the language."
     },
     {
       "icon": "community",
       "title": "Countless Packages",
-      "contents": "Dart allows users to build and publish their own packages for public use, which can be accessed from Dart directly."
+      "content": "Dart allows users to build and publish their own packages for public use, which can be accessed from Dart directly."
     },
     {
       "icon": "functional",
       "title": "Highly-organized Code",
-      "contents": "Dart's coding style makes it easy to read and understand."
+      "content": "Dart's coding style makes it easy to read and understand."
     },
     {
       "icon": "fun",
       "title": "Easy to Learn",
-      "contents": "Dart is easy to understand and great for beginners to programming."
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -618,7 +618,38 @@
       "name": "Type Conversion"
     }
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      icon: "web",
+      title: "Popular for Web and Mobile Development",
+      content: "Dart has the Flutter library, which can be used to quickly build web and smartphone applications."
```
Stargator, line 625: Web development can be a key feature as many libraries in Dart facilitate or support that. I would suggest dropping mention of Flutter here and instead use it as an example of the cross-platform feature.

If this section is for the web, then the language needs to focus on the web. Otherwise, this and the cross-platform sections are just repeating themselves.

config.json
```diff
@@ -618,7 +618,38 @@
...
+    },
+    {
+      icon: "community",
+      title: "Full Documentation",
+      contents: "Dart's documentation and learning materials are constantly updated by its developers, making it easy to look up information on the language."
+    },
+    {
+      icon: "community",
+      title: "Countless Packages",
+      contents: "Dart allows users to build and publish their own packages for public use, which can be accessed from Dart directly."
```
Stargator, line 640: This may be a good spot to add `pub.dev` where all the Dart packages live.

config.json
```diff
@@ -618,7 +618,38 @@
       "name": "Type Conversion"
     }
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      icon: "web",
```
Stargator, line 623: I would suggest the first listed feature to be that Dart is a general-purpose language that allows it to be flexible and used in a variety of use-cases.

config.json
```diff
@@ -618,7 +618,38 @@
...
+    {
+      icon: "functional",
+      title: "Highly-organized Code",
+      contents: "Dart's coding style makes it easy to read and understand."
+    },
+    {
+      icon: "fun",
+      title: "Easy to Learn",
+      contents: "Dart is easy to understand and great for beginners to programming."
+    }
```
Stargator, line 651: I would suggest adding a section on Dart's strong type system.

config.json
```diff
@@ -618,7 +618,38 @@
...
+    },
+    {
+      icon: "cross-platform",
+      title: "Instant iOS and Android Development",
+      content: "Dart/Flutter can be converted into iOS's Swift language and Android's Kotlin language, making mobile development fast and efficient."
+    },
+    {
+      icon: "community",
+      title: "Full Documentation",
+      contents: "Dart's documentation and learning materials are constantly updated by its developers, making it easy to look up information on the language."
```
Stargator, line 635: I think this is a great inclusion. This would be a good spot to point to `api.dart.dev` so folks can see the documentation in full.

Stargator (changes requested):
@andrewfgarrison Thanks for the update. I don't want us to conflate Flutter with Dart. So using Flutter as an example of Dart's cross-platform capabilities and robust design is good. I just don't want Flutter to over shadow Dart in the track's documentation.

I made some other suggestions, but this is not a do or fail situation. Feel free to bring up any questions about my feedback.

Thanks again.

config.json
Stargator: As an option, we can just collapse web and cross-platform into one section since web is part of cross-platform.

andrewfgarrison: I think combining web and cross-platform is fine, since all that is bundled together. In responses to your first comment, a single mention of Flutter in this section would probably be sufficient, avoiding mention of the library anywhere else.

config.json
andrewfgarrison: I thought about adding that, but I wasn't sure whether to mention strong typing exclusively, or Dart's flexibility in both strong and loose typing. Regardless, strong typing will most likely be the focus, but do you think mentioning the choice is worthwhile?

Stargator: Since Dart 2.0, the language has moved to a stronger typing system. You might not need to declare the type of something like `var statement = "Hello";`

But if you want to assign a number to that, Dart won't let you unless you explicitly change it to `dynamic statement = "Hello";`

So I would still say Dart has a strong type system even if you choose not to explicitly declare the type of everything.

config.json
Stargator: The syntax for JSON is for keys to be treated as strings. This would need to be resolved.

sk-dv:
how can I contribute to solve this pull request?

Stargator:
@andrewfgarrison Any interest in following up on this PR?

config.json
kytrinyx: The key features aren't meant purely to entice people into trying the language. We have only 100 characters to play with, so we'll have to leave the link to the api docs out.

kytrinyx:
I've updated this PR based on the feedback, and adjusted per `configlet lint` so that we can merge it.

config.json
kytrinyx: We already have 6 features in the config.json, and can't have more than that. I recommend merging as is, and we can later determine whether one of the features should be swapped out in favor of talking about strong typing.

config.json
kytrinyx: We have pretty tight character limits. The key features aren't really intended for directing people to resources; it's really just to entice people to try the track.

kytrinyx (approved):

Stargator (approved):

PR #371, opened 2021-12-05
links.yml: Check links for https availability

Stargator:
* Including --require-https flag as of version v1.1.1 will handle this

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -3,6 +3,7 @@ name: Links
 on:
   pull_request:
     paths:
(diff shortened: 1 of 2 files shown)


Stargator:
@devkabiir or @ErikSchierboom Could you please review this?

devkabiir (approved):

Stargator:
@devkabiir I will work on this some more tomorrow. Since the workflow for links was updated, the links workflow check should be run in the PR. So I will work on enabling that tomorrow

devkabiir:
The commit message is wrong on the last commit

PR #372, opened 2021-12-05
test.yml: Restrict action to only when dart related files change

Stargator:


Stargator:
@devkabiir or @ErikSchierboom Could you please review this?

.github/workflows/links.yml
```diff
@@ -2,6 +2,13 @@ name: Links
 
 on:
   pull_request:
+    paths:
+      - 'concept/**/*.md'
+      - 'concept/**/links.json'
+      - 'docs/*.md'
+      - 'exercises/**/*.md'
+      - 'reference/*.md'
+      - '*.md'
```
devkabiir, line 11: Perhaps `**/**/*.md` or `**/*.md` could be added as well to cover files other than top-level. Or does `*.md` already cover it?

.github/workflows/test.yml
```diff
@@ -1,6 +1,16 @@
...
 
-on: [pull_request]
+on:
+  pull_request:
+    paths:
+      - 'bin/*.dart'
+      - 'exercises/**/*.dart'
+      - 'exercises/**/analysis_options.yaml'
+      - 'exercises/**/pubspec.yaml'
+      - 'lib/src/*.dart'
```
devkabiir, line 10: `lib/*.dart` should also be considered.

devkabiir (approved):
If we happen to end up using any kind of fixtures in tests (e.g. from a json file) that dont match these glob patterns. The PR that updates the fixtures would silently pass, even if they dont pass the tests anymore.

.github/workflows/links.yml
Stargator: I thought it would, but it doesn't hurt to be more careful. We can always pare it back later.

.github/workflows/test.yml
Stargator: Good call

Stargator, line 10:
```suggestion
      - 'bin/*.dart'
      - 'exercises/**/*.dart'
      - 'exercises/**/analysis_options.yaml'
      - 'exercises/**/pubspec.yaml'
      - 'lib/*.dart'
      - 'lib/src/*.dart'
```

.github/workflows/links.yml
Stargator, line 11:
```suggestion
      - 'concept/**/*.md'
      - 'concept/**/links.json'
      - 'docs/*.md'
      - 'exercises/**/*.md'
      - 'reference/*.md'
      - '*.md'
      - '**/**/*.md'
      - '**/*.md'
```

PR #375, opened 2022-01-15
Do not skip tests for word-count excercise

synaptiko:
I noticed that tests are skipped. I think there is one more exercise which had the same problem but because I'm new to Dart I didn't realize it before.

### DIFF
--- a/exercises/practice/word-count/test/word_count_test.dart
+++ b/exercises/practice/word-count/test/word_count_test.dart
@@ -19,67 +19,67 @@ void simpleTests() {
   test('count one of each word', () {
     final Map<String, int> result = wordCount.countWords('one of each');
     expect(result, equals(<String, int>{'one': 1, 'of': 1, 'each': 1}));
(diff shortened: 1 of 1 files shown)


synaptiko:
I just found this https://github.com/exercism/dart/pull/347 but TBH, I wasn't aware there is such documentation. I followed the instructions on the website and I don't remember it would be mentioned anywhere. Maybe it would be good to add it to some visible place in the UI. Or maybe `exercism submit` command should report it to you (it should check that you don't have any skipped tests before the submit).

Stargator:
@synaptiko You can find the instructions on the right-hand side of the main Dart Track page. From there, you can click on [Testing on the Dart Track](https://exercism.org/docs/tracks/dart/tests).

Given some students confusion, we may consider adding it to the README for each exercise, but we would want to do that programmatically and not copy and paste.

If you'd like to help with that, feel free to do so. I'll create an issue so we can discuss it.

PR #385, opened 2022-04-12
Update hello_world_test.dart

kareemkhaalil:


Stargator:
@kareemkhaalil Thank you for finding your way here, unfortunately, the changes in this PR would break our test for the hello-world exercise. So I can't accept this change. But I encourage you to submit changes in the future if you see anything that can be improved.

PR #387, opened 2022-04-20
Update hello_world.dart

iBy3l:


Stargator:
@iBy3l Thank you for your contribution, unfortunately, this is the intended design of our Hello World problem. We have an implemented solution and the fix is just to change the string value. So we do not want to just give the users the solution, but we don't want them to have to create it from the ground up either.

Thank and if you see anything else that could be improved, we would welcome your submissions.

PR #390, opened 2022-05-01
solution

amrabdalfatah:
This is the solution for saying Hello World


Stargator:
Hey @amrabdalfatah, great job on the solution. But you don't need to submit solutions to the repository. You just use the `exercism` command to submit it so it gets tied to your Exercism account.

PR #392, opened 2022-05-21
Update hello_world.dart

karimelsherbinii:


Stargator:
@karimelsherbinii I think you meant to use `exercism submit` to send your solution to exercism.io. We do not accept solutions into the exercise as that would defeat the purpose.

PR #394, opened 2022-07-02
Update hamming_test.dart

jasper2virtual:
Both string input is empty, should be a error, instead of normal

### DIFF
--- a/exercises/practice/hamming/test/hamming_test.dart
+++ b/exercises/practice/hamming/test/hamming_test.dart
@@ -15,8 +15,7 @@ void main() {
 
   group('Hamming', () {
     test('empty strands', () {
-      final int result = hamming.distance('', '');
-      expect(result, equals(0));
+      expect(() => hamming.distance('', ''), throwsA(noEmptyStrand));
     }, skip: false);
 
     test('single letter identical strands', () {


Stargator:
@jasper2virtual Sorry for the wait/delay in getting to this PR. Thank you for your contribution.

You assert that an error should be thrown if one of the strings is empty. But from the instructions it states (emphasis mine):

> ### Implementation notes
> The Hamming distance is only defined for sequences of equal length, so an attempt to calculate it between sequences of different lengths should not work. **The general handling of this situation (e.g., raising an exception vs returning a special value) may differ between languages**.

So the instructions don't cover this specific case nor whether any invalid string should be an error.

However, upon reviewing the test suite we have these other examples using empty strings:

```dart
test('disallow left empty strand', () {
  expect(() => hamming.distance('', 'G'), throwsA(noEmptyStrand));
}, skip: true);

test('disallow right empty strand', () {
  expect(() => hamming.distance('G', ''), throwsA(noEmptyStrand));
}, skip: true);
```

So it seems like the first test is being treated differently than these latter two tests.

So I _could_ see the argument that for consistency the first test should result in an error. However, if properly implemented it _could_ render these two examples as being obsolete.

I'm going to need to review where Hamming currently resides in the order of exercises. That may help me consider this suggestion.

@jasper2virtual If you have anything else to add to the conversation, feel free to do so.

kytrinyx:
Another thing worth checking is what the current state of this exercise is in the upstream problem-specifications repository. While no track is required to follow what is there, it often has reasonable defaults.

https://github.com/exercism/problem-specifications/blob/main/exercises/hamming/canonical-data.json

Stargator:
@kytrinyx Thanks, the canonical has an empty strand expect `0` not an error. So I suggest @jasper2virtual if you want the empty strain to throw an error that you go to the https://github.com/exercism/problem-specifications and make your case there.

PR #395, opened 2022-07-18
Delete file two_fer_test.dart

aly22:


Stargator:
@aly22 What is the reasoning behind deleting this file?

I would think there is still value in having this practice exercise available to all users.

aly22:
Updating your old code. Try running it.On Sat, 30 Jul 2022, 21:00 Stargator, ***@***.***> ***@***.*** What is the reasoning behind deleting this file?I would think there is still value in having this practice exercise available to all users.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

aly22:
Updated your old code.

Stargator:
@aly22 Removing the required test file is not updating code, it'll result in the exercise no longer being functional for students. So I'm going to close this PR, unless you can provide a compelling reason.

PR #396, opened 2022-07-18
Update two_fer_test.dart

aly22:


Stargator:
@aly22 I believe you meant to use `exercism submit` to send the results you worked on to excersim.io

aly22:
Was just trying to update your old a** code...

Stargator:
@aly22 Updating the code by trying to add a file doesn't change the existing code. It just adds a second test file.

There's no need to insult the code, it's just text.

PR #404, opened 2022-10-20
Sync metadata for practice exercises

kytrinyx:
This runs

    bin/configlet sync --update --metadata --exercise SLUG

on any exercises that don't require any changes to their test suites in order to get up to date with the problem-specifications repository.

I generally prefer to update one exercise at a time, but reviewing gets quite tedious when there are lots of small updates, so I went with a bulk update here.

### DIFF
--- a/exercises/practice/leap/.meta/config.json
+++ b/exercises/practice/leap/.meta/config.json
@@ -19,5 +19,5 @@
   },
   "blurb": "Given a year, report if it is a leap year.",
   "source": "JavaRanch Cattle Drive, exercise 3",
-  "source_url": "https://coderanch.com/t/718816/Leap"
+  "source_url": "http://www.javaranch.com/leap.jsp"
 }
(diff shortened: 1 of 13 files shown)


devkabiir (approved):
LGTM 👍

exercises/practice/leap/.meta/config.json
```diff
@@ -19,5 +19,5 @@
   },
   "blurb": "Given a year, report if it is a leap year.",
   "source": "JavaRanch Cattle Drive, exercise 3",
-  "source_url": "https://coderanch.com/t/718816/Leap"
+  "source_url": "http://www.javaranch.com/leap.jsp"
```
Stargator, line 22: @kytrinyx A lot of these links should be updated to use https in the problem specifications. We had to fix them in order for our Links action to work. It's disabled temporarily due to an unrelated issue.

kytrinyx: Thanks for the heads up; I'll submit a PR to the problem-specifications repository.

Stargator: Thanks

PR #406, opened 2022-10-20
Sync space-age with problem-specifications

kytrinyx:
This updates the space-age exercise with changes
in the problem-specifications repository.

The instructions were updated with formatting changes.
The metadata updated a link to HTTP (rather than HTTPS) since the HTTPS version of the link doesn't work.
There was one additional test, that added an error case (invalid planet).

This exercise did not previously introduce error handling. It seems to me that error handling doesn't make the exercise more interesting. Also I think that keeping the exercises as simple as possible at the start is beneficial, and since there are other exercises that do introduce error handling, I don't think that the student misses out in the long run.

As such I chose not to bring the new test in here.

### DIFF
--- a/exercises/practice/space-age/.meta/config.json
+++ b/exercises/practice/space-age/.meta/config.json
@@ -18,5 +18,5 @@
   },
   "blurb": "Given an age in seconds, calculate how old someone is in terms of a given planet's solar years.",
   "source": "Partially inspired by Chapter 1 in Chris Pine's online Learn to Program tutorial.",
-  "source_url": "https://pine.fm/LearnToProgram/?Chapter=01"
+  "source_url": "http://pine.fm/LearnToProgram/?Chapter=01"
 }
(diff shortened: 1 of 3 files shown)


devkabiir (approved):

exercises/practice/space-age/.meta/config.json
```diff
@@ -18,5 +18,5 @@
   },
   "blurb": "Given an age in seconds, calculate how old someone is in terms of a given planet's solar years.",
   "source": "Partially inspired by Chapter 1 in Chris Pine's online Learn to Program tutorial.",
-  "source_url": "https://pine.fm/LearnToProgram/?Chapter=01"
```
Stargator, line 21: @kytrinyx This seems to be an issue with the source. The URL should be using HTTPS and our Link checker would otherwise fail it for not using HTTPS. We had disabled it because it failed in instances of mailto links.

I need to check if that's been fixed.

kytrinyx: I will fix this upstream and re-sync here.

PR #412, opened 2022-10-26
Sync allergies with problem-specifications

kytrinyx:
This puts all the tests in the same group rather than
grouping them per the canonical-data.json structure in
problem-specifications.

In order to help make the logical groupings apparent,
the test names have been updated to be prefixed with the
group description.

To make it as easy as possible to review, I've indented
the tests two spaces extra in the first commit, and then
fixed the indentation in the second commit.

In the third commit I've synced the allergies exercise by running 

```
bin/configlet sync --update --exercise allergies
```

This brought in updated instructions and an extra test.

### DIFF
--- a/exercises/practice/allergies/.meta/tests.toml
+++ b/exercises/practice/allergies/.meta/tests.toml
@@ -1,150 +1,160 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [17fc7296-2440-4ac4-ad7b-d07c321bc5a0]
-description = "not allergic to anything"
+description = "testing for eggs allergy -> not allergic to anything"
 
 [07ced27b-1da5-4c2e-8ae2-cb2791437546]
-description = "allergic only to eggs"
+description = "testing for eggs allergy -> allergic only to eggs"
 
 [5035b954-b6fa-4b9b-a487-dae69d8c5f96]
-description = "allergic to eggs and something else"
+description = "testing for eggs allergy -> allergic to eggs and something else"
 
 [64a6a83a-5723-4b5b-a896-663307403310]
-description = "allergic to something, but not eggs"
+description = "testing for eggs allergy -> allergic to something, but not eggs"
 
 [90c8f484-456b-41c4-82ba-2d08d93231c6]
-description = "allergic to everything"
+description = "testing for eggs allergy -> allergic to everything"
 
 [d266a59a-fccc-413b-ac53-d57cb1f0db9d]
-description = "not allergic to anything"
+description = "testing for peanuts allergy -> not allergic to anything"
 
 [ea210a98-860d-46b2-a5bf-50d8995b3f2a]
-description = "allergic only to peanuts"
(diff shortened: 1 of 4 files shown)


kytrinyx:
@devkabiir do you think this failure is happening because I'm running 2.18.3 locally? That's the latest stable version, and there's an open PR to update all the exercises to use 2.18, so I figured I might be able to get away with not figuring out how to run multiple versions of Dart locally.

kytrinyx:
@devkabiir Would you point me in the right direction here? I've run `dart format` on the file that is failing, but it seems like CI is checking for something different.

devkabiir:
By default `dart format` has a line length set to 80. We use 120 in the track. So the command you are looking for is `dart format -l 120 .`. Another way to check and fix auto-fixable things is to run `dart run bin/presubmit.dart`. That also runs all tests (all exercises) so it might take some time to complete, so I suggest the leaner `dart format -l 120 .`


I also noticed that we are using an outdated command in the `bin/presubmit` script. Which is another matter. And we have some exercises with outdated dart style issues, which can be fixed by running `dart format -l 120 --fix .`

devkabiir (approved):
- I understand the _Sync allergies with problem-specifications_ part of this PR.
- I don't understand why tests had to be grouped in a top-level group? Is it because of the missing `-` in the test case descriptions? as in
`list when: - just eggs` vs `list when: just eggs`

- If this is exercism wide change, would it mean all exercise test files have to be updated?

Stargator (dismissed):

exercises/practice/allergies/.meta/tests.toml
```diff
@@ -1,150 +1,160 @@
...
-description = "allergic to everything"
+description = "testing for eggs allergy -> allergic to everything"
 
 [d266a59a-fccc-413b-ac53-d57cb1f0db9d]
-description = "not allergic to anything"
+description = "testing for peanuts allergy -> not allergic to anything"
 
 [ea210a98-860d-46b2-a5bf-50d8995b3f2a]
-description = "allergic only to peanuts"
+description = "testing for peanuts allergy -> allergic only to peanuts"
```
Stargator, line 31: @devkabiir I do think a Dart plugin for the generator could try to catch instances of this and then create groups that are named using the text before the `->`

exercises/practice/allergies/test/allergies_test.dart
```diff
@@ -1,275 +1,259 @@
...
-                <String>['eggs', 'peanuts', 'shellfish', 'strawberries', 'tomatoes', 'chocolate', 'pollen', 'cats']));
-      }, skip: true);
-
-      test('no allergen score parts', () {
-        final List<String> result = allergies.list(509);
-        expect(
-            result, equals(<String>['eggs', 'shellfish', 'strawberries', 'tomatoes', 'chocolate', 'pollen', 'cats']));
-      }, skip: true);
-    });
+    test('testing for eggs allergy - not allergic to anything', () {
```
Stargator, line 8: I do like it the other way when there are groups of tests that could easily fall in the same batches.

But I just don't know if toml would be able to denote that.

kytrinyx: I'll implement grouping and regenerate the test suite, and we can go from there 👍

devkabiir (approved):
Let's merge this one now. We can iterate and fine tune the test generator in subsequent PRs. The main objective is to sync the exercise with problem spec. Which is accomplished.

kytrinyx:
I only did the first step, which is to regenerate the existing test suite.
When I went to do the second step and sync with problem-specifications I was reminded about the asterisks in the unordered list in the new markdown, so I went and submitted a PR to fix that in the problem-specifications repository.
https://github.com/exercism/problem-specifications/pull/2138

Once that is merged I'll do the sync and regenerate.

(I like to do it in two steps so that it's clear what changes the sync brings in.)

kytrinyx:
I've synced and regenerated.

This should be ready to go.

devkabiir (approved):

exercises/practice/allergies/.meta/tests.toml
devkabiir: The tests are generated from canonical-data.json. It already has some grouping. It should be feasible.

Issue #410, opened 2022-10-24
Structure of test files for generated exercises?


kytrinyx:
I've been working on a generator that I can use to update existing tests as well as generate stubs for new exercises for tracks.

The Dart track is popular on Exercism, and I'd like to spend some time adding new tests based on the problem-specifications.

I noticed that there are a few different ways of laying out tests in this repo, and I am planning on using the current structure of `word-count` as the basis for the generator:

```
import 'package:test/test.dart';
import 'package:word_count/word_count.dart';

final wordCount = WordCount();

void main() {
  group("WordCount: Simple Tests - ", simpleTests);
  group("WordCount: Ignore special characters - ", ignoreSpecialCharacters);
  group("WordCount: Works with numbers too - ", notJustWords);
  group("WordCount: Edge case - ", edgeCases);
}

void simpleTests() {
  // ...
}

void ignoreSpecialCharacters() {
  // ...
}

void notJustWords() {
  // ...
}

void edgeCases() {
  // ...
}
```

@Stargator @devkabiir Will this work for you, or is there another structure that you'd prefer?

Stargator:
I suggest looking at the create_exercise.dart under the bin directory.

It is used to create exercises based on problem specifications, though it may use some updates.

If it works generally well, then we could refactor it to update existing exercises.

kytrinyx:
Thanks!

Okay, so it looks like this is the structure that you'd like:
https://github.com/exercism/dart/blob/29df47067aeecc33c7e47775cc0f911ea9fcd5bf/bin/create_exercise.dart#L78-L90

Basically, all the tests in a single group, regardless of whether they are grouped differently in the problem-specifications `canonical-data.json` file, correct?

I am working on making a (mostly) language-agnostic generator that would only require a small language-specific add-on.
I've used it to update most of the exercises in the Ruby track, and generate exercises for the Chapel track. If you're okay with having a flat structure with all the tests in a single group, that simplifies things considerably.

devkabiir:
Ah!, the reason why I wasn't notified is because `@kabiir` is no longer my username. I don't have access to that account. This has happened quite a few times 😅 .

> If you're okay with having a flat structure with all the tests in a single group, that simplifies things considerably.

I would be okay _because_ it simplifies things. Since we can always iterate and as you mentioned a small language-specific add-on can help tune things to match the track's taste.
[...]

PR #414, opened 2022-10-28
Sync anagram with problem-specifications

kytrinyx:
- Regenerate anagram (Note that this exercise does not have groups in the canonical data.)
- Sync anagram with problem-specifications

### DIFF
--- a/exercises/practice/anagram/.meta/tests.toml
+++ b/exercises/practice/anagram/.meta/tests.toml
@@ -1,12 +1,24 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [dd40c4d2-3c8b-44e5-992a-f42b393ec373]
 description = "no matches"
 
 [b3cca662-f50a-489e-ae10-ab8290a09bdc]
 description = "detects two anagrams"
+include = false
+
+[03eb9bbe-8906-4ea0-84fa-ffe711b52c8b]
+description = "detects two anagrams"
+reimplements = "b3cca662-f50a-489e-ae10-ab8290a09bdc"
 
 [a27558ee-9ba0-4552-96b1-ecf665b06556]
 description = "does not detect anagram subsets"
@@ -40,6 +52,24 @@ description = "anagrams must use all letters exactly once"
 
 [85757361-4535-45fd-ac0e-3810d40debc1]
 description = "words are not anagrams of themselves (case-insensitive)"
+include = false
+
+[68934ed0-010b-4ef9-857a-20c9012d1ebf]
+description = "words are not anagrams of themselves"
+reimplements = "85757361-4535-45fd-ac0e-3810d40debc1"
+
+[589384f3-4c8a-4e7d-9edc-51c3e5f0c90e]
+description = "words are not anagrams of themselves even if letter case is partially different"
+reimplements = "85757361-4535-45fd-ac0e-3810d40debc1"
+
+[ba53e423-7e02-41ee-9ae2-71f91e6d18e6]
+description = "words are not anagrams of themselves even if letter case is completely different"
+reimplements = "85757361-4535-45fd-ac0e-3810d40debc1"
 
 [a0705568-628c-4b55-9798-82e4acde51ca]
 description = "words other than themselves can be anagrams"
+include = false
+
+[33d3f67e-fbb9-49d3-a90e-0beb00861da7]
+description = "words other than themselves can be anagrams"
+reimplements = "a0705568-628c-4b55-9798-82e4acde51ca"
--- a/exercises/practice/anagram/test/anagram_test.dart
+++ b/exercises/practice/anagram/test/anagram_test.dart
@@ -11,8 +11,8 @@ void main() {
     }, skip: false);
 
     test('detects two anagrams', () {
-      final List<String> result = anagram.findAnagrams('master', <String>['stream', 'pigeon', 'maters']);
-      expect(result, equals(<String>['stream', 'maters']));
+      final List<String> result = anagram.findAnagrams('solemn', <String>['lemons', 'cherry', 'melons']);
+      expect(result, equals(<String>['lemons', 'melons']));
     }, skip: true);
 
     test('does not detect anagram subsets', () {
@@ -66,13 +66,23 @@ void main() {
       expect(result, equals(<String>[]));
     }, skip: true);
 
-    test('words are not anagrams of themselves (case-insensitive)', () {
-      final List<String> result = anagram.findAnagrams('BANANA', <String>['BANANA', 'Banana', 'banana']);
+    test('words are not anagrams of themselves', () {
+      final List<String> result = anagram.findAnagrams('BANANA', <String>['BANANA']);
+      expect(result, equals(<String>[]));
+    }, skip: true);
+
+    test('words are not anagrams of themselves even if letter case is partially different', () {
+      final List<String> result = anagram.findAnagrams('BANANA', <String>['Banana']);
+      expect(result, equals(<String>[]));
+    }, skip: true);
+
+    test('words are not anagrams of themselves even if letter case is completely different', () {
+      final List<String> result = anagram.findAnagrams('BANANA', <String>['banana']);
       expect(result, equals(<String>[]));
     }, skip: true);
 
     test('words other than themselves can be anagrams', () {
-      final List<String> result = anagram.findAnagrams('LISTEN', <String>['Listen', 'Silent', 'LISTEN']);
+      final List<String> result = anagram.findAnagrams('LISTEN', <String>['LISTEN', 'Silent']);
       expect(result, equals(<String>['Silent']));
     }, skip: true);
   });
--- a/exercises/practice/anagram/test/anagram_test.dart
+++ b/exercises/practice/anagram/test/anagram_test.dart
@@ -4,92 +4,76 @@ import 'package:test/test.dart';
 final anagram = Anagram();
 
 void main() {
-  group("simple tests", simpleTests);
+  group('Anagram', () {
+    test('no matches', () {
+      final List<String> result = anagram.findAnagrams('diaper', <String>['hello', 'world', 'zombies', 'pants']);
+      expect(result, equals(<String>[]));
+    }, skip: false);
 
-  group("slightly less simple tests", slightlyLessSimpleTests);
+    test('detects two anagrams', () {
+      final List<String> result = anagram.findAnagrams('master', <String>['stream', 'pigeon', 'maters']);
+      expect(result, equals(<String>['stream', 'maters']));
+    }, skip: true);
 
-  group("more challenging tests", moreChallengingTests);
+    test('does not detect anagram subsets', () {
+      final List<String> result = anagram.findAnagrams('good', <String>['dog', 'goody']);
+      expect(result, equals(<String>[]));
+    }, skip: true);
 
-  /// Edge cases are scenarios that are problems that occur under specific conditions
-  /// Read more: https://en.wikipedia.org/wiki/Edge_case
-  group("edge case tests", edgeCaseTests);
-}
+    test('detects anagram', () {
+      final List<String> result = anagram.findAnagrams('listen', <String>['enlists', 'google', 'inlets', 'banana']);
+      expect(result, equals(<String>['inlets']));
+    }, skip: true);
 
-void simpleTests() {
-  test('no matches', () {
-    final List<String> result = anagram.findAnagrams('diaper', <String>['hello', 'world', 'zombies', 'pants']);
-    expect(result, equals(<String>[]));
-  }, skip: false);
-}
+    test('detects three anagrams', () {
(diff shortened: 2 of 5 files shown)


exercises/practice/anagram/test/anagram_test.dart
```diff
@@ -1,31 +1,21 @@
...
 void main() {
-  group("simple tests", simpleTests);
-
-  group("slightly less simple tests", slightlyLessSimpleTests);
-
-  group("more challenging tests", moreChallengingTests);
-
-  /// Edge cases are scenarios that are problems that occur under specific conditions
-  /// Read more: https://en.wikipedia.org/wiki/Edge_case
-  group("edge case tests", edgeCaseTests);
```
devkabiir, line 15: As one of the early exercises, I believe the purpose here was to inform the learner about Edge cases as they progress in the track/exercise. At some point it even made sense to me to distribute tests cases and group them in the order of their rareness. As this is beneficial in real-world software development. First you want to get the MVP out, then you want to flesh out all the bugs and edge cases.

As this was a hand-written exercise, it had the human element of thinking about the journey of the learner. 

I am curious to know what you think about automating this?

It removes the human element and makes it purely mechanical. Which is useful when you don't have enough human-resources.

exercises/practice/anagram/test/anagram_test.dart
```diff
@@ -1,31 +1,21 @@
...
+  group("Anagram: anagram tests - ", anagramTests);
```
devkabiir, line 7: A top-level group is enough for this exercise, as canonical-data does not have any grouping. We do not need to extract out test cases to a named function `anagramTests`, a closure would suffice here. But I can understand that this would be special casing in your test generator.

devkabiir (approved):
The PR accomplishes it's purpose. So its good to merge. I do have some open questions though.

exercises/practice/anagram/test/anagram_test.dart
kytrinyx: No, that's no problem, and a good shout. I'll tweak it so that it does it that way.

exercises/practice/anagram/test/anagram_test.dart
kytrinyx: I think that makes sense, if you'd like to open a PR that suggests the grouping in the `problem-specifications` (or start with an issue, maybe, in case the overwhelming majority don't want to go down that road, then you wouldn't have done the actual work up-front).

exercises/practice/anagram/test/anagram_test.dart
```diff
@@ -83,13 +69,23 @@ void edgeCaseTests() {
...
+    expect(result, equals(<String>[]));
+  }, skip: true);
+
+  test('words are not anagrams of themselves even if letter case is completely different', () {
+    final List<String> result = anagram.findAnagrams('BANANA', <String>['banana']);
     expect(result, equals(<String>[]));
   }, skip: true);
 
   test('words other than themselves can be anagrams', () {
-    final List<String> result = anagram.findAnagrams('LISTEN', <String>['Listen', 'Silent', 'LISTEN']);
```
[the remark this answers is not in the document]
kytrinyx: I'm not sure I understand the question. Anagrams are case-insensitive, and there are tests that specify this in the problem-specifications.

Were there case-sensitive tests in the dart track?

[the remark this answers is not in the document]
devkabiir: As evident from the following diffs, you can see that 'Listen', 'Banana', 'banana' were removed in the updated tests.

<details>
<summary>Diff 1</summary>

```diff
- final List<String> result = anagram.findAnagrams('LISTEN', <String>['Listen', 'Silent', 'LISTEN']);
+ final List<String> result = anagram.findAnagrams('LISTEN', <String>['LISTEN', 'Silent']);
```
</details>

<details>
<summary>Diff 2</summary>

```diff
-    final List<String> result = anagram.findAnagrams('BANANA', <String>['BANANA', 'Banana', 'banana']);
+    final List<String> result = anagram.findAnagrams('BANANA', <String>['BANANA']);
```
</details>

However, upon further inspection, I realized that 'Listen' was a redundant test case and the banana variants have been split into 3 separate test cases.

exercises/practice/anagram/test/anagram_test.dart
kytrinyx: > My question was more about the fact that by automating, you'd be removing (or at least reducing) the human creativity that each exercise can have.

I 100% agree that we want to introduce people to the language, its idioms, paradigms, etc.

The generator isn't something that is enforced. It's really just a tool, because for most exercises the design work is in deciding what the API to the subject under test is going to look like. The generator is fairly flexible, so we can always tweak what a test suite will look like. Or we can ignore the generator and hand-craft a test suite if it seems like that is better.

In the case of this test suite, there's nothing about the grouping that is inherently specific to Dart (to my mind, anyway). By adding grouping, people can benefit from the pedagogical onramp that you've thought out, or they can ignore it entirely and list out all the tests in a flat structure.

kytrinyx:
@devkabiir thanks for taking the time to discuss all the details.

It sounds like there are three avenues here:

1. Leave the test suite as it is, and bring new tests in manually.
2. Submit a pull request to the `problem-specifications` repository with grouping, and then use that to generate the test suite.
3. Generate the test suite with the flat structure.

Which approach would you prefer?

exercises/practice/anagram/test/anagram_test.dart
```diff
@@ -1,31 +1,21 @@
-import 'package:anagram/anagram.dart';
 import 'package:test/test.dart';
+import 'package:anagram/anagram.dart';
```
Stargator, line 2: The imports should be sorted alphabetically

Stargator (approved):

exercises/practice/anagram/test/anagram_test.dart
devkabiir: I noticed that too, but I purposefully did not mention it. I thought we should add a [lint rule](https://dart.dev/tools/linter-rules#directives_ordering) for that so `dart format --fix` takes care of it.

devkabiir:
> 1\. Leave the test suite as it is, and bring new tests in manually.

That sounds like more work for you. Future iterations of the problem spec. would require to do the same.

> 2\. Submit a pull request to the `problem-specifications` repository with grouping, and then use that to generate the test suite.

I might have been misleading. I sometimes do meta-commentary. Since we've been discussing #410 #412 #408 . I was talking in general about the grouping for tests in all exercises not just this one.
But I do understand that the way we've grouped tests in this track for `anagram` could benefit all.

> 3\. Generate the test suite with the flat structure.

More exercises/tracks will benefit from this as this will require your generator to consider such cases. Future sync/updates will be automatable as an effect.

I think the least effort, most fun & best outcome would be to go with option 3. 

Since you intend to polish things up in this track and bring it up to speed. Specifically _adding more exercises_. This is what concerned me about the test generator. When adding new exercises, generated tests would inevitably gravitate implementors towards a specific paradigm. But it's still better for the track than not having enough exercises.

exercises/practice/anagram/test/anagram_test.dart
devkabiir: Since this is a generated file, it's not trivial to know the ordering of these two imports and effectively sort them in a template. I think the template would need to add the aforementioned lint rule to `analysis_options.yaml` of the generated exercise and apply `dart format --fix`

kytrinyx: That is easy enough to do. Is there an example of where this is done correctly so that `dart format --fix` will work?

kytrinyx:
> I think the least effort, most fun & best outcome would be to go with option 3.

That works for me!

> Since you intend to polish things up in this track and bring it up to speed. Specifically adding more exercises. This is what concerned me about the test generator. When adding new exercises, generated tests would inevitably gravitate implementors towards a specific paradigm. 

In terms of adding exercises, I would be delighted to have help defining what a "typical" test would look like for a given exercise, and then generate based on that.

I can't do the design work, as I have never worked with Dart, so I wouldn't have good instincts. But if you're open to spending some time up front hashing out what the general API of the subject under test should be, etc, then I think that should be pretty simple to work with (for most exercises; there are some exceptions that I still don't have a good solution for).

Let's get the existing exercises brought up to speed first, and then once that's all good, I can open issues for a couple of exercises that we can discuss and then try out the generator with, and then iterate on the process from there.

exercises/practice/anagram/test/anagram_test.dart
kytrinyx: I tried adding `directives_ordering` to the linter rules in the `analysis_options.yaml` file that is in the `exercises/practice/anagram` directory, like so:

```
linter:
  rules:
    # Error Rules
    - avoid_relative_lib_imports
    - avoid_types_as_parameter_names
    - literal_only_boolean_expressions
    - no_adjacent_strings_in_list
    - valid_regexps
    
    # Style rules
    - directives_ordering
 ```

Then I ran `dart format --fix exercises/practice/anagram`, but it didn't change the ordering.

```
$ dart format --fix exercises/practice/anagram/
Formatted 2 files (0 changed) in 0.18 seconds.
```

Any suggestioons?

Comment (devkabiir) [reply exercises/practice/anagram/test/anagram_test.dart:2]: The lint rule `directives_ordering` has a quick fix instead of a style fix. Quick fixes are applied by `dart fix --apply`. Style fixes are applied by `dart format -l 120 .`. You can think of `dart format` as [prettier](https://prettier.io/) version of dart and `dart fix` as the [ESLint](https://eslint.org/) version of dart.

Comment (kytrinyx) [reply exercises/practice/anagram/test/anagram_test.dart:2]: Thank you, that worked! I'll push this up again.

File: exercises/practice/anagram/.meta/tests.toml [hunk 5]
```diff
@@ -40,6 +52,24 @@ description = "anagrams must use all letters exactly once"
...
+description = "words are not anagrams of themselves"
+reimplements = "85757361-4535-45fd-ac0e-3810d40debc1"
+
+[589384f3-4c8a-4e7d-9edc-51c3e5f0c90e]
+description = "words are not anagrams of themselves even if letter case is partially different"
+reimplements = "85757361-4535-45fd-ac0e-3810d40debc1"
+
+[ba53e423-7e02-41ee-9ae2-71f91e6d18e6]
+description = "words are not anagrams of themselves even if letter case is completely different"
+reimplements = "85757361-4535-45fd-ac0e-3810d40debc1"
```
devkabiir, line 67: Is this correct? 3 test cases `reimplement` the same test case

exercises/practice/anagram/.meta/tests.toml
```diff
@@ -1,12 +1,24 @@
...
 [dd40c4d2-3c8b-44e5-992a-f42b393ec373]
 description = "no matches"
 
 [b3cca662-f50a-489e-ae10-ab8290a09bdc]
 description = "detects two anagrams"
+include = false
+
+[03eb9bbe-8906-4ea0-84fa-ffe711b52c8b]
+description = "detects two anagrams"
+reimplements = "b3cca662-f50a-489e-ae10-ab8290a09bdc"
```
devkabiir, line 21: What does `reimplements` mean?

devkabiir (approved):

exercises/practice/anagram/.meta/tests.toml
kytrinyx: Yes, this is correct.
The TL;DR is that the original test was split into three tests for clarity.

See https://github.com/exercism/problem-specifications/pull/1943 for the full explanation.

"Reimplements" means that the new test specification supersedes the old one. Since test specifications are (largely) immutable, if you want to make a material change to a test case, you essentially need to deprecate the old one and implement a new one. "Reimplements" is the mechanism for this.

exercises/practice/anagram/.meta/tests.toml
kytrinyx: "Reimplements" means that the new test specification supersedes the old one. Since test specifications are (largely) immutable, if you want to make a material change to a test case, you essentially need to deprecate the old one and implement a new one (e.g. changing the inputs or expected values or both). "Reimplements" is the mechanism for this.

devkabiir: Thank you.

exercises/practice/anagram/test/anagram_test.dart
Stargator: @kytrinyx Spot on, if we can't assume someone has prior programming experience, then my intention was to introduce some broader concepts and terminology in the early exercises.

However, with the move to Exercism V3 these practice exercises can be tackled in any order, correct?

And we don't have a curriculum for the non-practice exercises, so assuming my understanding is correct, we'll have to rethink how we bring up these concepts to students.

I apologize for not getting the non-practice exercises created. I thought each track could set up its own path and I would work on the non-practice exercises every so often.

But during that time, changes kept being made to V3 about how things were done and I just got tired of having to refactor and change what I was doing.

So I got a little burned out. And now my time is pretty well spent during the week.

I do hope someone can sit up the non-practice exercises and keep some of these intentions on mine for the students who are coming fresh face into this track.

kytrinyx: > Spot on, if we can't assume someone has prior programming experience, then my intention was to introduce some broader concepts and terminology in the early exercises.

Got it. That makes a lot of sense.

Overall we do assume that people have prior programming experience if they're using Exercism (even if it's not a whole lot). In the early days of Exercism we really didn't understand who was using Exercism, but now it's clear that Exercism is a tool for learning programming languages, if you already know how to program (at least a little).

> However, with the move to Exercism V3 these practice exercises can be tackled in any order, correct?

Yes, that is true.

> I apologize for not getting the non-practice exercises created.

No apologies necessary! You have no obligation whatsoever to do that; I'm just grateful for all the work that you've put into this track.

> So I got a little burned out.

❤️ I completely understand!

Even if we do set up a syllabus for the Dart track, it will be focused on Dart-specific concepts rather than broader concepts within programming. We're starting to think about how to create courses that are specific to things other than programming languages, but it's a ways out yet.

PR #416, opened 2022-10-28
Sync docs with problem-specifications

kytrinyx:
This syncs the instructions for exercises that
do not have pending test changes.

The update ensures that urls use https rather than http, and
that unordered lists use a hyphen.


devkabiir (approved):
:rocket:

PR #419, opened 2022-10-30
Sync isbn-verifier with problem-specifications

kytrinyx:
- Regenerate isbn-verifier
- Sync isbn-verifier with problem-specifications

This brings in updated instructions as well as a couple
of new tests.


devkabiir (approved):
Thank you for your contribution. :rocket:

PR #421, opened 2022-10-30
Sync collatz-conjecture with problem specifications

kytrinyx:
This regenerate collatz-conjecture, and then syncs with problem-specifications.

The sync brought in updated instructions and
reimplemented test cases. The test cases changed the
expected error message from 'number' to 'integer' which
is more correct for the collatz-conjecture (which does
not apply to floats).

This means that existing solutions will fail the tests,
but with a failure that is trivial to fix.

### DIFF
--- a/exercises/practice/collatz-conjecture/.docs/instructions.md
+++ b/exercises/practice/collatz-conjecture/.docs/instructions.md
@@ -2,10 +2,11 @@
 
 The Collatz Conjecture or 3x+1 problem can be summarized as follows:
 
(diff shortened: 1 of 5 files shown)


Stargator (approved):
Thanks for the contribution.

devkabiir:
> This means that existing solutions will fail the tests,
> but with a failure that is trivial to fix.

You push merge the button 😝

kytrinyx:
😂 I take full responsibility.

Stargator:
That's perfectly fine. I come from the mindset that once it's been approved the author can merge it in as they may catch last minute issues.

PR #422, opened 2022-10-30
Sync hamming with problem-specifications

kytrinyx:
This brings in updated instructions as well as a couple
of new tests and several reimplemented tests that changed
the expected error message.

Since the error messages don't really make a material difference
I didn't change the implementation to use the new messages.

### DIFF
--- a/exercises/practice/hamming/.meta/lib/example.dart
+++ b/exercises/practice/hamming/.meta/lib/example.dart
@@ -1,11 +1,7 @@
 class Hamming {
   int distance(String a, String b) {
-    if ((a.isEmpty && b.isNotEmpty) || (a.isNotEmpty && b.isEmpty)) {
-      throw ArgumentError('no strand must be empty');
-    }
-
     if (a.length != b.length) {
-      throw ArgumentError('left and right strands must be of equal length');
+      throw ArgumentError('strands must be of equal length');
     }
 
     int diff = 0;
--- a/exercises/practice/hamming/test/hamming_test.dart
+++ b/exercises/practice/hamming/test/hamming_test.dart
@@ -4,11 +4,8 @@ import 'package:test/test.dart';
 void main() {
   final hamming = Hamming();
 
-  final mustEqual = predicate(
-      (Error e) => e is ArgumentError && e.message == 'left and right strands must be of equal length',
-      'an ArgumentError with the message "left and right strands must be of equal length"');
-  final noEmptyStrand = predicate((Error e) => e is ArgumentError && e.message == 'no strand must be empty',
-      'an ArgumentError with the message "no strand must be empty"');
+  final mustEqual = predicate((Error e) => e is ArgumentError && e.message == 'strands must be of equal length',
+      'an ArgumentError with the message "strands must be of equal length"');
 
   group('Hamming', () {
     test('empty strands', () {
@@ -44,12 +41,12 @@ void main() {
       expect(() => hamming.distance('ATA', 'AGTG'), throwsA(mustEqual));
     }, skip: true);
 
-    test('disallow left empty strand', () {
-      expect(() => hamming.distance('', 'G'), throwsA(noEmptyStrand));
+    test('disallow empty first strand', () {
+      expect(() => hamming.distance('', 'G'), throwsA(mustEqual));
     }, skip: true);
 
-    test('disallow right empty strand', () {
-      expect(() => hamming.distance('G', ''), throwsA(noEmptyStrand));
+    test('disallow empty second strand', () {
+      expect(() => hamming.distance('G', ''), throwsA(mustEqual));
     }, skip: true);
   });
 }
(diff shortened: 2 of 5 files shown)


exercises/practice/hamming/test/hamming_test.dart
```diff
@@ -47,12 +44,12 @@ void main() {
...
     }, skip: true);
 
-    test('disallow left empty strand', () {
-      expect(() => hamming.distance('', 'G'), throwsA(noEmptyStrand));
+    test('disallow empty first strand', () {
+      expect(() => hamming.distance('', 'G'), throwsA(mustEqual));
     }, skip: true);
 
-    test('disallow right empty strand', () {
-      expect(() => hamming.distance('G', ''), throwsA(noEmptyStrand));
```
Stargator, line 55: It seems noEmptyStrand is no longer used in this file. Unless I'm missing anything, you should be free to remove it.

Stargator (changes requested):

Stargator:
I see one of the hamming tests fail. I'm going to pull down the branch and look at how to resolve it so it throws error expected by the test.

kytrinyx:
I thought I had double-checked that they passed. I will also check on my end.

kytrinyx:
```
                                     
  Expected: throws an ArgumentError with the message "left and right strands must be of equal length"
    Actual: <Closure: () => int>
     Which: threw ArgumentError:<Invalid argument(s): no strand must be empty>
            stack package:hamming/hamming.dart 4:7  Hamming.distance
                  test/hamming_test.dart 48:28      main.<fn>.<fn>.<fn>
                  package:test_api                  expect
                  test/hamming_test.dart 48:7       main.<fn>.<fn>
  
  package:test_api             expect
  test/hamming_test.dart 48:7  main.<fn>.<fn>
  

To run this test again: dart test test/hamming_test.dart -p vm --plain-name 'Hamming disallow empty first strand'
00:01 +7 -1: test/hamming_test.dart: Hamming disallow empty second strand                                                                                         00:01 +7 -2: test/hamming_test.dart: Hamming disallow empty second strand [E]                                                                                                                          
  Expected: throws an ArgumentError with the message "left and right strands must be of equal length"
    Actual: <Closure: () => int>
     Which: threw ArgumentError:<Invalid argument(s): no strand must be empty>
            stack package:hamming/hamming.dart 4:7  Hamming.distance
                  test/hamming_test.dart 52:28      main.<fn>.<fn>.<fn>
                  package:test_api                  expect
                  test/hamming_test.dart 52:7       main.<fn>.<fn>
  
  package:test_api             expect
  test/hamming_test.dart 52:7  main.<fn>.<fn>
  
```

kytrinyx:
Ok, it looks like we can remove the first guard clause in the example solution. I'm pushing up a fix.

Stargator:
> Ok, it looks like we can remove the first guard clause in the example solution. I'm pushing up a fix.

Yeah, that's the solution. Verified locally.

exercises/practice/hamming/test/hamming_test.dart
Stargator: Verified it's been removed

Stargator (approved):

exercises/practice/hamming/.meta/lib/example.dart
```diff
@@ -1,9 +1,5 @@
 class Hamming {
   int distance(String a, String b) {
-    if ((a.isEmpty && b.isNotEmpty) || (a.isNotEmpty && b.isEmpty)) {
-      throw ArgumentError('no strand must be empty');
-    }
-
     if (a.length != b.length) {
       throw ArgumentError('left and right strands must be of equal length');
```
devkabiir, line 4: The language in testing seems to be changed from left/right to first/second? If that's intentional, we should be consistent here as well.

devkabiir (dismissed):
Thank you for your contribution :rocket:.  I have 1 concern.

exercises/practice/hamming/test/hamming_test.dart
```diff
@@ -1,17 +1,12 @@
...
-
-/// Test cases match one-to-one with those described
-/// in the problem specification found here:
-/// https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json
 void main() {
+  final hamming = Hamming();
+
   final mustEqual = predicate(
       (Error e) => e is ArgumentError && e.message == 'left and right strands must be of equal length',
       'an ArgumentError with the message "left and right strands must be of equal length"');
```
Stargator, line 9: Based on @devkabiir's other post, the strings will need to be update to reference "first" and "second" instead of left and right.

Stargator (changes requested):

exercises/practice/hamming/test/hamming_test.dart
kytrinyx: I updated to match what was in problem-specifications, which removes left/right altogether.

Stargator (approved):

Stargator:
@kytrinyx you are good to merge

PR #423, opened 2022-10-30
Sync isogram with problem-specifications

kytrinyx:
The first commit regenerates the test suite, which removes one level of nested grouping.

The second commit syncs with problem-specifications. Nothing exciting to see there!


Stargator (approved):

Stargator:
Thanks, @kytrinyx for all the contributions!

PR #424, opened 2022-10-30
Sync matching-brackets with problem specifications

kytrinyx:
This brings in improvements to the instructions as well
as some new tests.

The first commit regenerates the test suite in order to minimize the diff for the second commit, which does the actual sync.

### DIFF
--- a/exercises/practice/matching-brackets/test/matching_brackets_test.dart
+++ b/exercises/practice/matching-brackets/test/matching_brackets_test.dart
@@ -3,90 +3,91 @@ import 'package:test/test.dart';
 
 void main() {
   final MatchingBrackets matchingBrackets = new MatchingBrackets();
+
   group('MatchingBrackets', () {
-    test("paired square brackets", () {
-      final bool result = matchingBrackets.isPaired("[]");
+    test('paired square brackets', () {
+      final bool result = matchingBrackets.isPaired('[]');
       expect(result, equals(true));
     }, skip: false);
 
-    test("empty string", () {
-      final bool result = matchingBrackets.isPaired("");
+    test('empty string', () {
+      final bool result = matchingBrackets.isPaired('');
       expect(result, equals(true));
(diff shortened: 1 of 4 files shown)


exercises/practice/matching-brackets/test/matching_brackets_test.dart
```diff
@@ -3,90 +3,106 @@ import 'package:test/test.dart';
 
 void main() {
   final MatchingBrackets matchingBrackets = new MatchingBrackets();
+
   group('MatchingBrackets', () {
-    test("paired square brackets", () {
-      final bool result = matchingBrackets.isPaired("[]");
+    test('paired square brackets', () {
+      final bool result = matchingBrackets.isPaired('[]');
```
Stargator, line 9: @kytrinyx I've started going over some of my older branches for this repo that I haven't updated in awhile.

No change here, but Dart supports not using the type for variables. So in the future we could change it to:

```dart
final result = matchingBrackets.isPaired('[]');
```

Since Dart's a strongly typed language, it can infer that the variable is a bool based on the function's return type.

I'll create an issue to follow-up on this later.

Stargator (approved):

exercises/practice/matching-brackets/test/matching_brackets_test.dart
kytrinyx: That's great. Does that mean that we can simplify this as well?

```
final MatchingBrackets matchingBrackets = new MatchingBrackets();
```

devkabiir: > That's great. Does that mean that we can simplify this as well?
> 
> ```
> final MatchingBrackets matchingBrackets = new MatchingBrackets();
> ```

Yes, you can simplify that to

```dart
final matchingBrackets = MatchingBrackets();
```

PR #425, opened 2022-10-30
Sync prime-factors with problem-specifications

kytrinyx:
- Regenerate prime-factors to minimize diff when syncing
- Sync prime-factors with problem-specifications

This brings in improved instructions, ensures that the source url uses https
rather than http, and adds several new tests.


Stargator (approved):
Thanks, I see you've started taking in the changes regarding use of types for local variables. Thank you.

PR #427, opened 2022-10-30
 Sync resistor-color-duo with problem-specifications

kytrinyx:
- Regenerate resistor-color-duo in first commit to minimize diff when syncing
- Sync resistor-color-duo with problem-specifications

This brings in improved documentation and a couple of new tests.
It also removes the explicit type when assigning 'result' in the tests.


devkabiir (approved):
Thank you @kytrinyx 's test generator for your contribution :rocket:

PR #428, opened 2022-10-30
Sync triangle with problem-specifications

kytrinyx:
The first commit regenerates the test suite to minimize changes when doing the sync.

The sync brings in improvements to the instructions, ensures that
the source url in the metadata is https rather than http, and
adds a couple of tests.

### DIFF
--- a/exercises/practice/triangle/test/triangle_test.dart
+++ b/exercises/practice/triangle/test/triangle_test.dart
@@ -4,9 +4,9 @@ import 'package:triangle/triangle.dart';
 final triangle = Triangle();
 
 void main() {
-  group("Triangle: equilateral triangle - ", equilateralTriangle);
-  group("Triangle: isosceles triangle - ", isoscelesTriangle);
-  group("Triangle: scalene triangle - ", scaleneTriangle);
+  group('Triangle: equilateral triangle - ', equilateralTriangle);
+  group('Triangle: isosceles triangle - ', isoscelesTriangle);
+  group('Triangle: scalene triangle - ', scaleneTriangle);
 }
 
 void equilateralTriangle() {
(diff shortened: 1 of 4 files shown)


devkabiir (approved):
Whatever it is your test generator is having, I wanna have some of that too! 🍻

Stargator:
@kytrinyx any idea when we'll get a chance to look at your generator? I think @devkabiir  can't wait until he can see it for himself.

exercises/practice/triangle/test/triangle_test.dart
```diff
@@ -4,106 +4,118 @@ import 'package:triangle/triangle.dart';
...
-
-      test('sides may be floats', () {
-        final bool result = triangle.scalene(0.5, 0.4, 0.6);
-        expect(result, equals(true));
-      }, skip: true);
-    });
-  });
+  group("Triangle: equilateral triangle - ", equilateralTriangle);
+  group("Triangle: isosceles triangle - ", isoscelesTriangle);
+  group("Triangle: scalene triangle - ", scaleneTriangle);
```
Stargator, line 9: Our preference is to use single quotes when possible

Stargator (changes requested):

exercises/practice/triangle/test/triangle_test.dart
kytrinyx: Oh, yes! Something went awry here.  Let me see what happened.

kytrinyx: I have fixed and regenerated.

kytrinyx:
The generator is still very much an experiment. I'm constantly messing with it to figure out what it needs to do. I've used it for a handful of tracks and have locked down the basic functionality I will probably do a clean rewrite and make it publicly available. I don't think it would be useful for anyone to look at all the thrashing that is happening right now.

Stargator (approved):

Issue #426, opened 2022-10-30
Drop use of types for variables when not necessary


Stargator:
Dart supports not using the type for variables. So we could change it from:

```dart
final bool result = matchingBrackets.isPaired('[]');
```

To:

```dart
final result = matchingBrackets.isPaired('[]');
```

Since Dart's a strongly typed language, it can infer that the variable is a bool based on the function's return type.

Stargator:
#429 Closes this

PR #429, opened 2022-10-31
Remove explicit type definitions

kytrinyx:
Per #426, drop types for variable definitions when assigning
results in test suites.

This removes type definitions for test suites that have
been synced with problem-specifications recently.


Stargator (changes requested):
Can you also remove the use of `new` for the constructors? They are also optional in Dart.

kytrinyx:
Yes! Good call.

Stargator (approved):

PR #431, opened 2022-11-02
Sync word-count with problem-specifications

kytrinyx:
The first commit regenerates the word-count exercise based on the previously selected test cases, in order to minimize the diff when syncing.

Note that the problem-specification did not group the tests, so regenerating it removed the manually-created grouping.

The second commit syncs with problem-specifications. This brought in two new tests. The example solution had to be tweaked to pass the updated test suite.

<details>
<summary>Failed test</summary>

```
00:00 +8: test/word_count_test.dart: WordCount with apostrophes                                                                                                   00:00 +8 -1: test/word_count_test.dart: WordCount with apostrophes [E]                                                                                                                                 
  Expected: {
              'first': 1,
              'don\'t': 2,
              'laugh': 1,
              'then': 1,
              'cry': 1,
              'you\'re': 1,
              'getting': 1,
              'it': 1
            }
    Actual: {
              'first': 1,
              'don\'t': 2,
              'laugh': 1,
              'then': 1,
              'cry': 1,
              'you\'r': 1,
              'e': 1,
              'getting': 1,
              'it': 1
            }
     Which: has different length and is missing map key 'you\'re'
  
  package:test_api                expect
  test/word_count_test.dart 50:7  main.<fn>.<fn>
```

</details>

### DIFF
--- a/exercises/practice/word-count/.docs/instructions.md
+++ b/exercises/practice/word-count/.docs/instructions.md
@@ -12,7 +12,7 @@ When counting words you can assume the following rules:
 
 1. The count is _case insensitive_ (ie "You", "you", and "YOU" are 3 uses of the same word)
 2. The count is _unordered_; the tests will ignore how words and counts are ordered
-3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are ignored
+3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are regarded as spaces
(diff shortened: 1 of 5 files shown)


Stargator:
@kytrinyx Thanks. That failed yesterday is funny since it's counting the `e` as a separate word.

I'll look at it soon (a few hours), but not sure I'll have a quick turnaround on the solution.

Stargator (approved):

Stargator:
Looks like you fixed the test, looks good. Approved

PR #432, opened 2022-11-03
Sync luhn from problem-specifications

kytrinyx:
The first commit manually updates strings to use single-quotes rather than double-quotes, in order to minimize the diff when regenerating. The second commit regenerates the test suite based on the existing `tests.toml` file, which surprisingly added new tests. The third commit syncs with problem-specifications, which adds even more tests. None of the tests had any surprises, though; just more edge cases. The existing example solution didn't need to be updated at all.


devkabiir (approved):
Thank you for your contribution. 🚀

PR #433, opened 2022-11-03
Sync gigasecond with problem-specifications

kytrinyx:
The first commit regenerates the test suite. I decided to remove the three first tests, as they are all redundant,
keeping only the two that are specified in the problem-specifications
using full datetimes as opposed to just dates.

The second commit syncs with problem-specifications. The sync updated the instructions.
There was an additional test that I chose not to include, as it
is a test for immutability, but as far as I can tell, the Dart
DateTime type is immutable already, so the test doesn't make
sense here.

### DIFF
--- a/exercises/practice/gigasecond/test/gigasecond_test.dart
+++ b/exercises/practice/gigasecond/test/gigasecond_test.dart
@@ -2,49 +2,19 @@ import 'package:gigasecond/gigasecond.dart';
 import 'package:test/test.dart';
 
 void main() {
-  group("Gigasecond", gigasecondTests);
-}
-
-void gigasecondTests() {
-  group('Add one gigasecond to the input. ', () {
-    test('date only specification of time', () {
-      final DateTime birthDate = DateTime.utc(2011, DateTime.april, 25);
-      final DateTime result = add(birthDate);
-      DateTime expectedDate = DateTime.utc(2043, DateTime.january, 1, 1, 46, 40);
-
-      expect(result, equals(expectedDate));
+  group('Gigasecond', () {
+    test('full time specified', () {
+      final moment = DateTime.utc(2015, DateTime.january, 24, 22, 00, 00);
+      final result = add(moment);
+      final expected = DateTime.utc(2046, DateTime.october, 02, 23, 46, 40);
+      expect(result, equals(expected));
     }, skip: false);
 
-    test('second test for date only specification of time', () {
-      final DateTime birthDate = DateTime.utc(1977, DateTime.june, 13);
-      final DateTime result = add(birthDate);
-      DateTime expectedDate = DateTime.utc(2009, DateTime.february, 19, 1, 46, 40);
-
-      expect(result, equals(expectedDate));
-    }, skip: true);
-
-    test('third test for date only specification of time', () {
-      final DateTime birthDate = DateTime.utc(1959, DateTime.july, 19);
-      final DateTime result = add(birthDate);
-      DateTime expectedDate = DateTime.utc(1991, DateTime.march, 27, 1, 46, 40);
-
-      expect(result, equals(expectedDate));
-    }, skip: true);
-
-    test('full time specified', () {
-      final DateTime birthDate = DateTime.utc(2015, DateTime.january, 24, 22, 00, 00);
-      final DateTime result = add(birthDate);
-      DateTime expectedDate = DateTime.utc(2046, DateTime.october, 2, 23, 46, 40);
-
-      expect(result, equals(expectedDate));
-    }, skip: true);
-
     test('full time with day roll-over', () {
-      final DateTime birthDate = DateTime.utc(2015, DateTime.january, 24, 23, 59, 59);
-      final DateTime result = add(birthDate);
-      DateTime expectedDate = DateTime.utc(2046, DateTime.october, 3, 01, 46, 39);
-
-      expect(result, equals(expectedDate));
+      final moment = DateTime.utc(2015, DateTime.january, 24, 23, 59, 59);
+      final result = add(moment);
+      final expected = DateTime.utc(2046, DateTime.october, 03, 01, 46, 39);
+      expect(result, equals(expected));
     }, skip: true);
   });
 }
--- a/exercises/practice/gigasecond/test/gigasecond_test.dart
+++ b/exercises/practice/gigasecond/test/gigasecond_test.dart
@@ -5,14 +5,14 @@ void main() {
   group('Gigasecond', () {
     test('full time specified', () {
       final moment = DateTime.utc(2015, DateTime.january, 24, 22, 00, 00);
-      final result = add(moment);
+      final result = addGigasecondTo(moment);
       final expected = DateTime.utc(2046, DateTime.october, 02, 23, 46, 40);
       expect(result, equals(expected));
     }, skip: false);
 
     test('full time with day roll-over', () {
       final moment = DateTime.utc(2015, DateTime.january, 24, 23, 59, 59);
-      final result = add(moment);
+      final result = addGigasecondTo(moment);
       final expected = DateTime.utc(2046, DateTime.october, 03, 01, 46, 39);
       expect(result, equals(expected));
     }, skip: true);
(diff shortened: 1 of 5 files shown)


exercises/practice/gigasecond/test/gigasecond_test.dart
```diff
@@ -2,49 +2,19 @@ import 'package:gigasecond/gigasecond.dart';
...
-    }, skip: true);
-
-    test('third test for date only specification of time', () {
-      final DateTime birthDate = DateTime.utc(1959, DateTime.july, 19);
-      final DateTime result = add(birthDate);
-      DateTime expectedDate = DateTime.utc(1991, DateTime.march, 27, 1, 46, 40);
-
-      expect(result, equals(expectedDate));
-    }, skip: true);
-
```
devkabiir, line 33: I would argue that `date`, `date-time` are hard concepts to wrap your brain around. Every single tool/language does it differently. There are standards which have inconsistencies in them and then when they are implemented they are not standards compliant.

While `gigasecond` is not specifically about `date`, `date-time`.  And while the goal of an exercise is more about being fun to solve. 

I think we should keep these tests as they are specifically trying to make sure the correctness of `gigasecond` _in Dart_.

What kind of shenanigans does Dart's date, date-time implementation have? I don't remember. But would I want my date, date-time related code to work as expected in a real world? You bet!.

Since dart can be compiled to web (JavaScript), does it affect result of this `gigasecond` implementation? Does Dart compile it's own implementation of DateTime or does it use the one provided by browser implementation? I don't really care!. I mostly care that it works as expected.

I think the date only tests could get a better description than just saying `second test...` or `third test...`, that would've helped justify their existence better.


P.S. Although it may seem like it, my tone is _not_ sarcastic. It's more playful, but after reading my comment I deemed it necessary to clarify.

exercises/practice/gigasecond/test/gigasecond_test.dart
```diff
@@ -2,49 +2,19 @@ import 'package:gigasecond/gigasecond.dart';
...
 
 void main() {
-  group("Gigasecond", gigasecondTests);
-}
-
-void gigasecondTests() {
-  group('Add one gigasecond to the input. ', () {
-    test('date only specification of time', () {
-      final DateTime birthDate = DateTime.utc(2011, DateTime.april, 25);
-      final DateTime result = add(birthDate);
```
devkabiir, line 12: The name of the `add` function is very generic. It might break tests for the existing solutions, but I think a more appropriate name would be `addGigaSecond` or `addGigaSecondTo`.

devkabiir (changes requested):

devkabiir:
> There was an additional test that I chose not to include, as it
> is a test for immutability, but as far as I can tell, the Dart
> DateTime type is immutable already, so the test doesn't make
> sense here.

You are correct. But do the learners know that?.  Date, DateTime and Immutability are hard concepts. Especially for new learners.
These concepts are also very _relative_ in nature. Their meaning and schematics changes based on _what_ (language/framework/platform) you're talking about, _who_ you're talking to and _when_. 

The test checks that the `add` function does not mutate the _input_. There is no guarantee that the input will be the base `DateTime` object. 

In some programming languages, date-time is a primitive that cannot be subclassed.  But in dart, one can subclass `DateTime`, which means it can be made mutable. Often, in real world software development, people will use what's familiar to them, and as such, they might use a date-time library to do date-time operations. That library might return a subclass of DateTime that's mutable. Whether this mutability is a bug or a feature is not the point here.

```dart
class MyDate extends DateTime {
    int day;
    MyDate(int year, int month, this.day): super.utc(year, month, day);
}

void main() {
    final mydate = MyDate(2022, 10, 3);
    print(mydate.day);

    /// The day can not only be changed, but can also be incorrect depending on the implementation
    mydate.day = 999999;
    print(mydate.day);
}
```

I would say we have some options here,
- Add some hints about (im)mutability of DateTime in Dart.
- Implement the test regardless of the hints.

kytrinyx:
That's useful to know. I will go ahead and implement the test.

(Side note, it would be interesting to have add "approaches" to the new dig deeper tab for this exercise. Here's the PR that adds approaches for the gigasecond exercise in the C# track: https://github.com/exercism/csharp/pull/2015/files)

exercises/practice/gigasecond/test/gigasecond_test.dart
kytrinyx: I like `addGigaSecondTo`. Very explicit and nice! I'll go ahead and change that.

exercises/practice/gigasecond/test/gigasecond_test.dart
kytrinyx: I see where you're coming from, but the tests don't seem to accomplish any of this. They simply define a `DateTime` with no hours, minutes, or seconds. Does this really help explore the problem space?

Perhaps we could invent a new exercise that really does accomplish the above?

kytrinyx:
I've pushed an update that implements the immutability check. I had to guess a bit what it should look like, please let me know if there's a better way to express this.

exercises/practice/gigasecond/test/gigasecond_test.dart
devkabiir: > Does this really help explore the problem space?

I thought about. I could not convince myself. So my answer is no.

> Perhaps we could invent a new exercise that really does accomplish the above?

I think that's the better approach, since `gigasecond` is really a one-liner exercise. One of the easy ones. Does Exercism have exercises with the theme of "common pitfalls"? This could fall under that.

devkabiir:
> I've pushed an update that implements the immutability check. I had to guess a bit what it should look like, please let me know if there's a better way to express this.

I don't see any tests?

I also don't know how to properly test this. One would need to allow mutability first and be able to inspect the input afterwards for any changes.
This would require to implement the whole interface of the `DateTime` class and allow mutability for fields like `second`, `day`, etc.
It is hard to justify the test efforts for a one-liner exercise.

But then again, we are talking about `DateTime` and immutability here. One could argue it's justified

After trying locally to test this in various ways, I have realized that  I was the one at fault here!

Following is the definition of the function, that one shall implement. The input is restricted to `DateTime`
```dart
DateTime addGigasecondTo(DateTime moment) {}
```

What this essentially means is that I would have to know at implementation time that there is a subclass `MyDate` that the end-user could potentially pass.
I would have to know the symbol name of that class and know that this subclass allows mutability.

Which is not really possible. What about `MyAwesomeDate` from a different package? How would I know that?

Since I am restricted to the interface of `DateTime`. I would not be able to do `moment.day = 9999`. This would be a type error.

If I really knew about the mutable class, I would need to be able to import it and refer to it and cast to it.
`(moment as MyDate).day = 999`. This would work, but again, requires me to know about the universe and all the unknowns.


So your initial decision of not including the test was brilliant!

devkabiir (approved):
Thank you for your contributions :rocket:

devkabiir:
> (Side note, it would be interesting to have add "approaches" to the new dig deeper tab for this exercise. Here's the PR that adds approaches for the gigasecond exercise in the C# track: https://github.com/exercism/csharp/pull/2015/files)

Thanks for informing about the new dig deeper tab. I've had enough `DateTime`s and numbers for the day. 😫

exercises/practice/gigasecond/test/gigasecond_test.dart
kytrinyx: Ok, excellent. We don't have concept-exercises that are "common pitfalls" as such, but there certainly are ways of addressing common pitfalls as part of exercises.

If you are feeling inspired to tackle something, go ahead and open an issue about it and we can figure out how to narrow it down to an actual exercise.

kytrinyx:
> I don't see any tests?

Ah. That will teach me to push up a commit just as I'm headed into a meeting. I apparently forgot to commit the new test, so the pushed code only had the tweaks to the test interface.

> So your initial decision of not including the test was brilliant!

That is a most excellent conclusion 😂 

Ok, cool! So I'm going to go ahead and merge this. We can always come back to the exercise later if we think of an interesting way to extend it or tweak it.

PR #435, opened 2022-11-03
Sync phone-number with problem-specifications

kytrinyx:
The first commit regenerates the test suite. This removed nesting, which I believe was done manually.

The second commit syncs with problem-specifications, updating the instructions and reimplementing two tests to get less ambiguous error situations.

### DIFF
--- a/exercises/practice/phone-number/test/phone_number_test.dart
+++ b/exercises/practice/phone-number/test/phone_number_test.dart
@@ -1,128 +1,120 @@
 import 'package:phone_number/phone_number.dart';
 import 'package:test/test.dart';
 
-final phoneNumber = PhoneNumber();
-
 void main() {
-  group('PhoneNumber: Cleanup special characters - ', cleanUpTest);
-  group('PhoneNumber: Number length - ', numberLengthTest);
-  group('PhoneNumber: Accept only numbers - ', numbersOnlyTest);
-  group('PhoneNumber: Area/Exchange Code - ', areaCodeTests);
-  group('PhoneNumber: Edge cases - ', exchangeCodeTests);
-}
+  final phoneNumber = PhoneNumber();
 
-void cleanUpTest() {
-  test('cleans the number', () {
-    final String? result = phoneNumber.clean('(223) 456-7890');
-    expect(result, equals('2234567890'));
-  }, skip: false);
+  group('PhoneNumber', () {
+    test('cleans the number', () {
+      final result = phoneNumber.clean('(223) 456-7890');
+      expect(result, equals('2234567890'));
+    }, skip: false);
 
-  test('cleans numbers with dots', () {
-    final String? result = phoneNumber.clean('223.456.7890');
-    expect(result, equals('2234567890'));
-  }, skip: true);
+    test('cleans numbers with dots', () {
+      final result = phoneNumber.clean('223.456.7890');
+      expect(result, equals('2234567890'));
+    }, skip: true);
 
-  test('cleans numbers with multiple spaces', () {
-    final String? result = phoneNumber.clean('223 456   7890   ');
-    expect(result, equals('2234567890'));
-  }, skip: true);
-}
+    test('cleans numbers with multiple spaces', () {
+      final result = phoneNumber.clean('223 456   7890   ');
+      expect(result, equals('2234567890'));
+    }, skip: true);
 
-void numberLengthTest() {
-  test('invalid when 9 digits', () {
-    expect(() => phoneNumber.clean('123456789'),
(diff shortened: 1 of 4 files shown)


exercises/practice/phone-number/test/phone_number_test.dart
```diff
@@ -1,128 +1,120 @@
...
+          () => phoneNumber.clean('123456789'),
+          throwsA(
+              predicate<FormatException>((e) => e is FormatException && e.message == 'incorrect number of digits')));
+    }, skip: true);
+
+    test('invalid when 11 digits does not start with a 1', () {
+      expect(
+          () => phoneNumber.clean('22234567890'),
+          throwsA(
+              predicate<FormatException>((e) => e is FormatException && e.message == '11 digits must start with 1')));
```
devkabiir, line 34: One can write the test predicate in a less verbose manner without loosing any type safety at all

```dart
predicate((FormatException e) => e.message == '11 digits must start with 1')
```

```dart
predicate<FormatException>((e) => e.message == '11 digits must start with 1')
```

```dart
predicate((e) => e is FormatException && e.message == '11 digits must start with 1')
```


The first two make is impossible to throw an error that is other than `FormatException`, when doing so, the test will fail with an error that gives a hint like `which is not an <Instance of FormatException>`

<details>
<summary>Test Output Example</summary>

```log
  Expected: throws satisfies function
    Actual: <Closure: () => String?>
     Which: threw _Exception:<Exception: 11 digits must start with 1>
            stack package:phone_number/example.dart 17:7  PhoneNumber.clean
                  test/phone_number_test.dart 65:30       numbersOnlyTest.<fn>.<fn>
                  package:test_api                        expect
                  test/phone_number_test.dart 65:5        numbersOnlyTest.<fn>
                  
            which not an <Instance of 'FormatException'>
  
  package:test_api                  expect
```


</details>


The last option allows any object to be thrown as long as it is `FormatException` or a subclass of it. The syntax also makes it more clear that the error needs to be a `FormatException`. When this predicate fails the test will fail with an error that points to the predicate in the test file

I prefer the fist one.

Notice that in the test output we have the line `Expected: throws satisfies function` which is correct since we use a `predicate` as an assertion. We can make it more clear by providing a second argument `description`, which describes the intention of the predicate like so,

```dart
predicate(
      (FormatException e) => e.message == '11 digits must start with 1',
      'FormatException with message "11 digits must start with 1"'),
)
```
Notice that I have a trailing comma after the last argument, to make formatting easier.

The full test code would be 
```dart
test('invalid when 11 digits does not start with a 1', () {
  expect(() => phoneNumber.clean('22234567890'),
    throwsA(
      predicate(
        (FormatException e) => e.message == '11 digits must start with 1',
        'FormatException with message "11 digits must start with 1"',
      ),
    ),
  );
}, skip: true);
```

Depending on the number of trailing commas you add, you can reduce the indentation (more readable) or end up making it unreadable.


There is a simpler way to accomplish all of the above, by not using `predicate`. Instead use the `isA` matcher for assertion which suites our purpose better and has precise error output in test failure. We don't need to explicitly specify and `descriptions` for the assertion failure.

```dart
test('invalid when 11 digits does not start with a 1', () {
  expect(() => phoneNumber.clean('22234567890'),
    throwsA(
      isA<FormatException>().having(
        (e) => e.message,
        'message',
        '11 digits must start with 1',
      ),
    ),
  );
}, skip: true);
```

<details>
<summary>Test Output using isA matcher</summary>

```log
  Expected: throws <Instance of 'FormatException'> with `message`: '11 digits must start with 1'
    Actual: <Closure: () => String?>
     Which: threw _Exception:<Exception: 11 digits must start with 1>
            stack package:phone_number/example.dart 29:9  PhoneNumber.clean
                  test/phone_number_test.dart 40:30       numberLengthTest.<fn>.<fn>
                  package:test_api                        expect
                  test/phone_number_test.dart 40:5        numberLengthTest.<fn>
                  
            which is not an instance of 'FormatException'
```

</details>

Specifically the first line gives enough hint to the learner.
```
Expected: throws <Instance of 'FormatException'> with `message`: '11 digits must start with 1'
```

devkabiir (approved):
I have some comments for improving the test suite. As far as he purpose of this PR is concerned, it is accomplished. Thus good to merge.

exercises/practice/phone-number/test/phone_number_test.dart
kytrinyx: I've tweaked the test output to follow the style described here.

Would you take a look at it to make sure I've not missed anything?

devkabiir (approved):
Great contribution!

PR #437, opened 2022-11-04
Sync armstrong-numbers with problem-specifications

### LINKED ISSUE #434: Armstrong Numbers: do we want to add BigInt?
The armstrong-numbers exercise has two new tests in problem-specifications.

<details>
<summary>New test specifications</summary>

```
{
  "uuid": "5ee2fdf8-334e-4a46-bb8d-e5c19c02c148",
  "description": "Armstrong number containing seven zeroes",
  "comments": [
    "The numeric size of this input number is 108 bits. Consider skipping this test if appropriate."
  ],
  "scenarios": [
    "big-integer"
  ],
  "property": "isArmstrongNumber",
  "input": {
    "number": 186709961001538790100634132976990
  },
  "expected": true
}

{
  "uuid": "12ffbf10-307a-434e-b4ad-c925680e1dd4",
  "description": "The largest and last Armstrong number",
  "comments": [
    "The numeric size of this input number is 127 bits. Consider skipping this test if appropriate."
  ],
  "scenarios": [
    "big-integer"
  ],
  "property": "isArmstrongNumber",
  "input": {
    "number": 115132219018763992565095597973971522401
  },
  "expected": true
}
```

</details>

If I'm not mistaken this would require handling the new cases with `BigInt`.

Do you think this would be a good addition for this track?

Right now `armstrong-numbers` is listed as difficulty 1 (scale is 1-10). Would adding BigInt complicate matters enough that we would have to bump the difficulty up a bit? Do we want to do that?

I don't have a strong opinion, but at the moment I'm leaning slightly in favor of leaving the exercise as is. I'm happy to be convinced otherwise, though.

@Stargator @devkabiir Thoughts?

kytrinyx:
This updates the instructions and brings in two new tests,
both of which handle very big numbers.

In order to support these new tests, the example solution
has been updated to use BigInt.

Closes #434

### DIFF
--- a/exercises/practice/armstrong-numbers/.meta/lib/example.dart
+++ b/exercises/practice/armstrong-numbers/.meta/lib/example.dart
@@ -3,17 +3,16 @@ import 'dart:math' show pow;
 class ArmstrongNumbers {
   /// The parameters and variables within the method that are set to final in order to prevent the accidentally modification of the value.
   /// As those variables are not needed to be changed.
-  bool isArmstrongNumber(final num input) {
-    final String numberAsString = input.toString();
-    final int numOfDigits = numberAsString.length;
-    num sum = 0;
+  bool isArmstrongNumber(final String input) {
+    final numOfDigits = input.length;
+    var sum = BigInt.from(0);
 
-    for (int count = 0; count < numOfDigits; count++) {
-      final String digitAsString = numberAsString.substring(count, count + 1);
-      final int digit = int.parse(digitAsString);
-      sum += pow(digit, numOfDigits);
+    for (var count = 0; count < numOfDigits; count++) {
+      final digitAsString = input.substring(count, count + 1);
+      final digit = BigInt.parse(digitAsString);
+      sum += digit.pow(numOfDigits);
     }
 
-    return input == sum;
+    return input == sum.toString();
   }
 }
--- a/exercises/practice/armstrong-numbers/.meta/lib/example.dart
+++ b/exercises/practice/armstrong-numbers/.meta/lib/example.dart
@@ -1,5 +1,3 @@
-import 'dart:math' show pow;
-
 class ArmstrongNumbers {
   /// The parameters and variables within the method that are set to final in order to prevent the accidentally modification of the value.
   /// As those variables are not needed to be changed.
--- a/config.json
+++ b/config.json
@@ -132,7 +132,7 @@
         "uuid": "4f650701-9568-4d95-9f72-be4bde2c90ec",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 1,
+        "difficulty": 2,
         "topics": [
           "equality",
           "integers",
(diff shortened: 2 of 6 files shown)


exercises/practice/armstrong-numbers/.meta/lib/example.dart
```diff
@@ -3,17 +3,16 @@ import 'dart:math' show pow;
 class ArmstrongNumbers {
```
devkabiir, line 3: The import at the top of this file can be removed
```diff
-    import 'dart:math' show pow;
```

exercises/practice/armstrong-numbers/.meta/lib/example.dart
```diff
@@ -3,17 +3,16 @@ import 'dart:math' show pow;
...
-      final int digit = int.parse(digitAsString);
-      sum += pow(digit, numOfDigits);
+    for (var count = 0; count < numOfDigits; count++) {
+      final digitAsString = input.substring(count, count + 1);
+      final digit = BigInt.parse(digitAsString);
+      sum += digit.pow(numOfDigits);
     }
 
-    return input == sum;
+    return input == sum.toString();
```
devkabiir, line 16: Since you mentioned _approaches_ in #433 

What do you think of an approach which uses `int` for small enough numbers & switches `BigInt` for big ones?
- This could be accomplished by checking the input length. My rational behind this approach is that `BigInt`s are generally very very slow compared to `int`. 
- So if you were to build a program that deals with Money, or numbers in general & your numbers are generally small enough to fit in 64 bits. 
- This approach would make your program fast for most of the common use cases and daily needs. 
- And once a while when your program needs to deal with very large numbers, it switches to `BigInt`s, the performance will be noticeably slow for certain calculations but the end-users would expect that, since they have some understanding that big numbers are being crunched.

While not necessarily related to `armstrong-numbers`. I think this could benefit the learners.
I am curious to know whats your take on this?

devkabiir (approved):
Thank you for your contribution 🤖 . Merge at will.

exercises/practice/armstrong-numbers/.meta/lib/example.dart
kytrinyx: Good catch, thanks!

exercises/practice/armstrong-numbers/.meta/lib/example.dart
kytrinyx: Yepp. I think it's a great example of an interesting approach that would be worth exploring in a dig deeper article. Ruby does something similar with strings. The underlying C implementation uses one implementation for shorter strings and another for longer strings, if I recall correctly, for similar reasons.

Here's an example of approaches explored in C# for the "bob" exercise, including an additional article about performance trade-offs between different solutions: https://exercism.org/tracks/csharp/exercises/bob/dig_deeper (you might have to look at it in a private browser tab; not sure).

PR #452, opened 2022-11-06
Simplify assignment in diamond

kytrinyx:
This removes the explicit assignments in diamond,
allowing the types to be inferred.

### DIFF
--- a/exercises/practice/diamond/test/diamond_test.dart
+++ b/exercises/practice/diamond/test/diamond_test.dart
@@ -1,120 +1,87 @@
 import 'package:diamond/diamond.dart';
 import 'package:test/test.dart';
 
-final diamond = Diamond();
-
 void main() {
+  final diamond = Diamond();
+
   group('Diamond', () {
-    test('Degenerate case with a single "A" row', () {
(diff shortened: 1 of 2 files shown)


exercises/practice/diamond/test/diamond_test.dart
```diff
@@ -1,120 +1,87 @@
...
-      expect(
-        result,
-        equals([
-          ' A ',
-          'B B',
-          ' A ',
-        ]),
-      );
+      final result = diamond.rows('B');
+      expect(result, equals(<String>[' A ', 'B B', ' A ']));
```
Stargator, line 15: @kytrinyx Do note we have the list within `equals` formatted in a way so it's displayed as a diamond. Since the exercise is about generating these diamond string, I think we would like to maintain that. So it does require manual change.

Stargator (changes requested):

exercises/practice/diamond/test/diamond_test.dart
kytrinyx: Good call. I've updated these to have the original diamond shape.

PR #453, opened 2022-11-07
Add dart analyzer to Test workflow

Stargator:
We should include the Dart analyzer when we run the workflow. Though it only checks the bin, lib, and test folders because the exercises can't be checked. The exercises would throw errors because they are referencing a file that does not have the expected implementations.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,5 +34,9 @@ jobs:
     - name: Check Stage formatting
       run: dart run bin/check_formatting.dart
 
+    # We can't check the exercises because they will produce errors since the example files have the solution.
+    - name: Analyze Dart code
+      run: dart analyze bin/ lib/ test/
+
     - name: Run tests
       run: dart test


kytrinyx (approved):
This looks good to me.

If we did want to check the exercises, we could potentially copy the example file over the solution stub before running the analysis, assuming that this all happens in a tmp directory somewhere, it probably wouldn't matter that the git repo is left in a dirty state.

Stargator:
> This looks good to me.
> 
> If we did want to check the exercises, we could potentially copy the example file over the solution stub before running the analysis, assuming that this all happens in a tmp directory somewhere, it probably wouldn't matter that the git repo is left in a dirty state.

Yeah, this is what the exercise_test does for tests. Either we rename that script and have it do tests and analysis (not a good practice to do more than one thing) or generalize the common logic and create a new script that reuses some of it for `dart analyze `

kytrinyx:
That makes sense. Either way I think it makes sense to get this merged now, and leave those changes for a potential later improvement.

Stargator:
> That makes sense. Either way I think it makes sense to get this merged now, and leave those changes for a potential later improvement.

Agreed. Merging

PR #460, opened 2022-11-07
Regenerate robot-simulator

kytrinyx:
This regenerates the test suite for robot-simulator.
In doing so it:
- removes the explicit type when assigning the robot
- moves the custom expected() definition to the top of the file
- moves the group definitions out of main

I'm not sure if moving the expected() definition violates dart
style. In the generator I have custom initialization at the top,
and this felt like it was basically the same thing, so I didn't
introduce another way of customizing.

If you have strong feelings about it, I can change it.

### DIFF
--- a/exercises/practice/robot-simulator/.meta/config.json
+++ b/exercises/practice/robot-simulator/.meta/config.json
@@ -3,7 +3,8 @@
     "Corentin-Leffy"
   ],
   "contributors": [
-    "Stargator"
+    "Stargator",
+    "kytrinyx"
   ],
   "files": {
     "solution": [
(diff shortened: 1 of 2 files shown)


Stargator:
> I'm not sure if moving the expected() definition violates dart
> style. In the generator I have custom initialization at the top,
> and this felt like it was basically the same thing, so I didn't
> introduce another way of customizing.

The Dart style guide does not get into that level of detail, in terms of the order of public, private, or customize logic.

My only concern is pushing the `main()` down to the bottom. it's a central part of a dart app so I'm find with some global properties/references being defined as long as it does not obscure the main().

I don't think this PR does that, so i have no issue.

Stargator (approved):

PR #461, opened 2022-11-08
Sync forth with problem-specifications

kytrinyx:
The tests.toml file was missing for this exercise,
so I added it and regenerated the test suite.

The sync brought in improved instructions, but from
what I can tell there were no new tests added since the
exercise was implemented, though one test name was changed:
- before: cannot redefine numbers
- now: cannot redefine non-negative numbers

The set-up is now happening in each test rather than globally.
Also the generator pulled the test group definitions out of
main, which reduced the level of nesting.


exercises/practice/forth/.meta/config.json
```diff
@@ -16,7 +17,5 @@
       ".meta/lib/example.dart"
     ]
   },
-  "blurb": "Implement an evaluator for a very simple subset of Forth.",
-  "source": "Exercism JS track",
```
Stargator, line 20: Isn't there some way to mention when an exercise stems from another track?

exercises/practice/forth/test/forth_test.dart
```diff
@@ -2,294 +2,391 @@ import 'package:forth/forth.dart';
...
+    var forth = Forth();
+    forth.evaluate('1 2 dup');
+    expect(forth.stack, equals(<int>[1, 2, 2]));
+  }, skip: true);
+
+  test('errors if there is nothing on the stack', () {
+    var forth = Forth();
+    expect(
+      () => forth.evaluate('dup'),
+      throwsA(isA<Exception>().having((e) => e.toString(), 'message', 'Exception: Stack empty')),
```
Stargator, line 174: The error handling is a bit duplicative. Can we extract it into its own property and then reuse that property?

Same would go for any other error handling that is repeatedly used.

Stargator (changes requested):

exercises/practice/forth/.meta/config.json
kytrinyx: There is! I totally missed that this got deleted. I will replace it with the forked from entry that we settled on.

exercises/practice/forth/test/forth_test.dart
kytrinyx: Definitely!

exercises/practice/forth/.meta/config.json
kytrinyx: I was wrong. We don't have this for practice exercises, as they're assumed to be implemented based on problem-specifications, or made up from whole cloth.

We may consider introducing authors and contributors in problem-specifications as well, but we haven't done so yet.

Stargator: Ok, that may have been what I was thinking of. Thanks

Stargator (approved):

exercises/practice/forth/test/forth_test.dart
kytrinyx: I've re-generated and pushed the branch.

PR #465, opened 2022-11-14
Add stub for new exercise: high-scores

kytrinyx:
This is a proposal for a new exercise based on the canonical data in the problem-specifications repository.

I didn't make any design choices whatsoever—if you think a different API would be better suited to this exercise, please suggest some changes, and I will regenerate the test suite.

I put myself as contributor rather than author, since I don't intend to write the example solution.
As such, the authorship of this exercise is up for grabs to anyone who wants to spend a few minutes writing some code that will pass the tests.

### DIFF
--- a/exercises/practice/high-scores/.meta/config.json
+++ b/exercises/practice/high-scores/.meta/config.json
@@ -1,0 +1,19 @@
+{
+  "authors": [],
+  "contributors": [
+    "kytrinyx"
+  ],
+  "files": {
+    "solution": [
+      "lib/high_scores.dart"
+    ],
+    "test": [
+      "test/high_scores_test.dart"
+    ],
+    "example": [
+      ".meta/lib/example.dart"
+    ]
+  },
+  "blurb": "Manage a player's High Score list.",
+  "source": "Tribute to the eighties' arcade game Frogger"
+}
--- a/exercises/practice/high-scores/.meta/config.json
+++ b/exercises/practice/high-scores/.meta/config.json
@@ -1,5 +1,7 @@
 {
-  "authors": [],
+  "authors": [
+    "Stargator"
+  ],
   "contributors": [
     "kytrinyx"
   ],
--- a/exercises/practice/high-scores/.meta/lib/example.dart
+++ b/exercises/practice/high-scores/.meta/lib/example.dart
@@ -1,0 +1,55 @@
+class HighScores {
+
+  List<int> scores(List<int> inputScores) {
+    return inputScores;
+  }
+
+  int latest(List<int> inputScores) {
+    return inputScores.last;
+  }
+
+  int personalBest(List<int> inputScores) {
+    inputScores.sort();
+    return inputScores.reversed.first;
+  }
+
+  List<int> personalTopThree(List<int> inputScores) {
+    inputScores.sort();
+    final sortedScores = inputScores.reversed.toList(growable: false);
+    var result = <int>[];
+
+    if (inputScores.length >= 3) {
+      result = [
+        sortedScores.first,
+        sortedScores[1],
+        sortedScores[2]
+      ];
+    } else if (inputScores.length == 2){
+      result = [
+        sortedScores.first,
+        sortedScores[1]
+      ];
+    } else {
+      result = [
+        sortedScores.first
+      ];
+    }
+    return result;
+  }
+
+  int latestAfterTopThree(List<int> inputScores) {
+    return inputScores.last;
+  }
+
+  List<int> scoresAfterTopThree(List<int> inputScores) {
+    return inputScores;
+  }
+
+  int latestAfterBest(List<int> inputScores) {
+    return inputScores.last;
+  }
+
+  List<int> scoresAfterBest(List<int> inputScores) {
+    return inputScores;
+  }
+}
--- a/exercises/practice/high-scores/.meta/lib/example.dart
+++ b/exercises/practice/high-scores/.meta/lib/example.dart
@@ -36,20 +36,4 @@ class HighScores {
     }
     return result;
   }
-
-  int latestAfterTopThree(List<int> inputScores) {
-    return inputScores.last;
-  }
-
-  List<int> scoresAfterTopThree(List<int> inputScores) {
-    return inputScores;
-  }
-
-  int latestAfterBest(List<int> inputScores) {
-    return inputScores.last;
-  }
-
-  List<int> scoresAfterBest(List<int> inputScores) {
-    return inputScores;
-  }
 }
--- a/exercises/practice/high-scores/.meta/lib/example.dart
+++ b/exercises/practice/high-scores/.meta/lib/example.dart
@@ -1,7 +1,7 @@
 class HighScores {
   final List<int> _scores;
 
-  const HighScores(this._scores);
+  HighScores(this._scores);
 
   List<int> get getScores {
     return this._scores;
--- a/exercises/practice/high-scores/.meta/lib/example.dart
+++ b/exercises/practice/high-scores/.meta/lib/example.dart
@@ -1,37 +1,35 @@
 class HighScores {
-  List<int> scores = [];
+  final List<int> _scores;
 
-  HighScores(List<int> inputScores) {
-    this.scores = inputScores;
-  }
+  const HighScores(this._scores);
 
   List<int> get getScores {
-    return scores;
+    return this._scores;
   }
 
-  int latest() {
-    return scores.last;
+  int get latest {
+    return this._scores.last;
   }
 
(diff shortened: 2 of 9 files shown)


Stargator:
Thanks @kytrinyx . I'll take this on to complete the example and define the initial class.

Stargator:
@kytrinyx The high scores exercise is missing the pubspec.yaml file. I can generate it, just letting you know if you want it make it part of your generator.

Stargator:
@kytrinyx  I've implemented a solution for high-scores, it's very basic and just does the bare minimal to pass the tests.

What I noticed is a few instances where I didn't even have to implement the intended logic, like for the following methods:
- latest

- latestAfterTopThree

- scoresAfterTopThree

- latestAfterBest

- scoresAfterBest

I didn't need to actually solve the problem. I think the problem spec is missing some other cases that would push the user to really solve the intended problem.

kytrinyx:
For `latest` you didn't have to say `.last` in order to pass the test?

I think I've completely misunderstood the canonical-spec. Those tests aren't actually supposed to be methods; they're checking that one doesn't mutate the structures when calling the methods that are mentioned.

I will overhaul this.

kytrinyx:
Okay, I looked more closely and it doesn't make sense to implement those test cases, as they're supposed to be immutability checks. I've updated `tests.toml` to not include those cases, regenerated the test suite, and updated the example solution to remove those extra properties.

Stargator:
> I think I've completely misunderstood the canonical-spec. Those tests aren't actually supposed to be methods; they're checking that one doesn't mutate the structures when calling the methods that are mentioned.
> 
> I will overhaul this.

@kytrinyx Dart does do immutability. So if the HighScores class is suppose to have these properties then we can implement them.

I think what is expected to happen is the HighScores has a `scores` method that takes in the input. Then the properties are used to get the desired output.

The current test suite is passing the inputs into methods named after the properties, instead of the scores method. That's how I _could_ read the canonical-data.json

kytrinyx:
@Stargator I've updated the exercise. Please feel free to edit the example solution in whatever way you see fit.

If you'd like me to tweak the test suite to lead people towards a more idiomatic solution, please let me know what to change.

Stargator:
@kytrinyx I'll take a look

Stargator:
@kytrinyx, this spec is just confusing to me.

I don't understand the logic for some of these:

- latestAfterTopThree
  - If input is `[10, 20, 30, 40]`, I would expect the value to be `10`. Instead it's `20`.
  - The top three are 40, 30, and 20. So I would have thought 10 would be _after_ the top three.

- scoresAfterTopThree
  - If the input is `[0, 5, 10, 20, 30, 40]`, I would expect the value to be `[10, 5, 0]`. Instead, it's the same as the input.
  - The top three are 40, 30, and 20. So if we want a list of **scores** _after_ the top three, it would be all the scores except for 40, 30, and 20.

- latestAfterBest
  - This works as I would expect.

- scoresAfterBest
  - If the input is `[20, 30, 40]`, I would expect the value to be `[30, 20]`. Instead, it's the same as the input.
  - The best is `40`. So if we want a list of **scores** _after_ the best, it would be all the scores except for 40.

Stargator:
I've updated the test suite and example with what'll it take to implement immutability for the HighScores class. [Set the constructor to const](https://dart.dev/tools/diagnostic-messages?utm_source=dartdev&utm_medium=redir&utm_id=diagcode&utm_content=const_initialized_with_non_constant_value#const_with_non_const).

For more info:
* [Language Tour ](https://dart.dev/guides/language/language-tour#constructors)(search for "constant constructors")
* [Effective Dart section for constructors](https://dart.dev/guides/language/effective-dart/design#consider-making-your-constructor-const-if-the-class-supports-it)

kytrinyx:
> latestAfterTopThree

Latest after top three checks that if you call personalTopThree, and then call latest, latest still returns the same result as before.

Same with the others.

It doesn't mean that the class itself is immutable, it means that calling `personalTopThree` doesn't cause other methods to break.

ErikSchierboom:
> It doesn't mean that the class itself is immutable, it means that calling personalTopThree doesn't cause other methods to break.

@Stargator This test was added for tracks where it is a common mistake to do something like this (in pseudocode):

```
pub fn personalTopThree {
  this.scores.sort()
  this.scores.last(3)
}
```

This method does return the top three scores, but it has _also_ mutated its state, which means that a call to `latest()` might have a different result (unless the last score was also the highest score).

What each track needs to decide is whether having these tests makes sense for that track. For the C# track, I _did_ choose to implement these tests, as the C# track's implementation is designed to use state to keep track of the scores: https://github.com/exercism/csharp/blob/main/exercises/practice/high-scores/HighScoresTests.cs
For the F# track though, the test cases are all just regular functions, no state involved, so I did not add these tests: https://github.com/exercism/fsharp/blob/main/exercises/practice/high-scores/HighScoresTests.fs
What the Dart track should do is decide if they want these scores to be kept via state or not. 
If so, implementing the tests makes sense, otherwise, I'd omit them.

Stargator:
Thanks, @kytrinyx and @ErikSchierboom, I read your responses earlier, but didn't have the time until now to sit down and re-assess the PR in light of it.

Stargator:
@kytrinyx Okay, I've removed the `const` from the constructor and updated the test suite accordingly. Tests are passing locally.

The biggest differences are in the example, which isn't generated.

Though in the test suite, since the problem specification refers to `latest` and the like as properties, I changed the example.dart to have them seen as properties and not as methods.

If it's better for the generator for them to be methods, we can switch them back. Just trying to have the test suite adhere to the spec.

ErikSchierboom:
> Though in the test suite, since the problem specification refers to latest and the like as properties, I changed the example.dart to have them seen as properties and not as methods.

Problem specifications has no concept of properties vs methods, there are just property names. It is up to tracks to determine whether they want to implement a property as a function/method/property. So basically, all the property name is saying is that there is something you can test that can be described via a certain property name, but how that is implemented is up to the track.

kytrinyx:
> all the property name is saying is that there is something you can test that can be described via a certain property name

To riff on this: the property name doesn't have to match what you call things in the track. It's just a way of describing the current situation.

For example `allergies` has a property name `isAllergicTo` whereas the idiomatic way to express this in Ruby is `allergic_to?`.

Sometimes tracks depart quite a lot from the property name given in the problem-specifications data. It's not wrong, it just depends on what you're trying to express and how it makes sense to express it in that language.

kytrinyx:
@Stargator I've gone back to an earlier version here, where the tests expect methods rather than properties.

Would you take a look at this and evaluate whether this test suite makes sense from a Dart point of view?

exercises/practice/high-scores/.meta/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "authors": [],
+  "contributors": [
+    "kytrinyx"
```
Stargator, line 4: Could you please add my name there as well, since I created the solution?

Stargator (approved):

exercises/practice/high-scores/.meta/config.json
kytrinyx: Yes, as with the other exercise I will add you as `author`.

exercises/practice/high-scores/.meta/lib/example.dart
```diff
@@ -0,0 +1,46 @@
+class HighScores {
+  List<int> scores = [];
```
Stargator, line 2:
```suggestion
  List<int> scores = [];
```

I would have all references to scores or this.scores, changed to `_scores` or `this._scores`. An underscore makes the property `private` and we want that otherwise, the getter `getScores` does not make any sense.

Because the user/dev can pull the score values and redefine them at any time.

exercises/practice/high-scores/.meta/lib/example.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  HighScores(List<int> inputScores) {
+    this.scores = inputScores;
+  }
+
+  List<int> get getScores {
```
Stargator, line 8: In Dart, [when you define a `getter`, you don't include `get` in the name.](https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get)

`getter`s are generally used  (and explicitly in this case) to allow private properties to be access without altering their values.

Stargator (changes requested):

Stargator:
@kytrinyx Looks good to me.

Stargator (approved):

kytrinyx:
Thanks, @Stargator. And thanks for taking the time to work through this with me.

PR #489, opened 2023-04-11
sync sum of multiples docs

kytrinyx:
- Sync sum-of-multiples docs with problem-specifications
- Sync sum-of-multiples docs with problem-specifications


kytrinyx:
Yepp, pretty sure that's what happened. I've now updated my script to make sure I start from a clean base /forheadslap

This should be good now.

PR #529, opened 2024-02-02
actions: Update action references to use versions instead of hashes

Stargator:
This should reduce the number of Dependabot PRs that are created and reviewed


glennj:
I though that Exercism is trying to use hashes instead of versions. @ErikSchierboom am I misunderstanding

iHiD:
Yeah, we deliberately use hashes as versions risk introducing section risks, as someone can republish a version with a different, dangerous commit. It's worth noting this has actually previously happened, so it definitely feels like a real risk.

However, I do wonder if for "official" GitHub actions such as `checkout`, we can rely on the versions.

iHiD:
>  Also, shouldn't PRs have enough checks to verify before we update an action that it's working properly?

I think the issue is that actions can read secrets. And so if someone maliciously changes an action upstream and then we inadvertently run it, we leak things.

---

I'll leave @ErikSchierboom to answer your other questions as he knows more! :)

Stargator:
Closing this due to Exercism policy

Stargator:
> You could consider changing the dependabot frequency.

I don't have permissions to change the Dependabot settings for this repo

ErikSchierboom:
You do though, as it is configured in https://github.com/exercism/dart/blob/main/.github/dependabot.yml (ping me in the PR)

PR #581, opened 2024-07-12
pubspec: Upgrade sdk constraints to Dart 3.4 as minimum

fapdash:
See https://github.com/exercism/dart-test-runner/pull/84 for associated test runner change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tag: ['3.2']
+        tag: ['3.4']
 
     container:
       image:  dart:${{ matrix.tag }}
--- a/exercises/concept/futures/pubspec.yaml
+++ b/exercises/concept/futures/pubspec.yaml
@@ -1,5 +1,5 @@
 name: 'futures'
 environment:
-  sdk: '>=3.2.0 <4.0.0'
+  sdk: '>=3.4.0 <4.0.0'
 dev_dependencies:
   test: '<2.0.0'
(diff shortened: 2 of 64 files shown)


fapdash:
I looked through the changelog of the 3.3 and 3.4 releases and the potential for breakage of student solutions seems low, but not zero. See https://github.com/dart-lang/sdk/issues/54056

glennj:
v3.2 is only a few months old. It seems premature to drop it. What's your motivation for this change?

fapdash:
@glennj I wanted this repo to match the test runner after https://github.com/exercism/dart-test-runner/pull/84.
Should I only change the Dart version the GHA is running on?

glennj:
OK, but why are you suggesting upversioning the test runner? What's the rationale behind it all?

fapdash:
Two reasons: 
1. I saw that https://github.com/exercism/dart/commit/71baf2a1ce7ea3bca73a4324f6282edc6878ef64 changed the dependency versions as well, which I now realize might be related to the Dart major versions making that necessary.
2. I like to run on the latest and greatest. :D

I'm fine with any changes you suggest to this PR, my goal was to make the track run on the latest stable Dart release before I start diving into the track as a student.

glennj:
I'd suggest you work on the exercises first, and if things aren't working like you expect then we can look at upgrading.

Stargator:
@fapdash, please note the last update took us from 2.18 to 3.2. So we were behind on supporting the current major stable version of Dart.

Dart 3.4 can still be used by students because our current setup  (`sdk: '>=3.2.0 <4.0.0'`) supports it.

When Dart 4.0 comes out, we'll likely to wait until 4.1 or 4.2 before switching to let things settle down.

I'm not entirely against this change, but per @glennj's guidance, I think it the impacts on students should be weighed against the rewards.

There's nothing stopping students from using Dart 3.3 or 3.4, or a future 3.x. Unless there's an exercise that is not working or the test runner can't function without an upgrade, there doesn't seem to be a good reason to update.

fapdash:
Sure thing, closing then.

PR #586, opened 2024-09-22
Add atbash-cipher

BNAndras:


BNAndras:
Run locally, I don't see any changes to be made.

```
(base) ➜  atbash-cipher git:(add-atbash-cipher) dart format .
Formatted 2 files (0 changed) in 0.09 seconds.
```

As an aside, shouldn't a formatting check be separate from the testing of the solution? At the moment, it's a prerequisite before we test the example, but it might not have any bearing on the example. What if the formatting issue was in the stub, but not the example? I'd recommend either making a separate workflow for the formatting check (I've seen a few tracks do this) or make the formatter run after the tests pass.

glennj:
Yeah, that's a good idea. Can you make an issue out of it?

glennj (approved):

BNAndras:
See #591

PR #588, opened 2024-09-22
Add `nucleotide-count`

BNAndras:


config.json
```diff
@@ -172,6 +172,14 @@
           "math"
         ]
       },
+      {
+        "slug": "nucleotide-count",
+        "name": "Nucleotide Count",
+        "uuid": "06ea660b-3869-46df-8fd3-a0651b66ed54",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2
```
glennj, line 181: This is the first exercise where the user has to throw an error. They'll probably need to research that a bit. I'd suggest bumping the difficulty a notch, but leave it in this order.

glennj (approved):

config.json
BNAndras: I can do an instructions append like Python does to briefly introduce the concept of errors and point students to resources like the Dart docs

glennj: That would be fantastic.

BNAndras: Still working on the instructions.append file, but for now, I'm switching to using exceptions. I figure that should be a little simpler for the reader to grasp conceptually. In my reading, an exception boils down to "something unexpected happened, but it's not the programmer's fault". A string containing something besides `A`, `C`, `G`, and `T` fits that because the caller provided something invalid when our code only handles those four letters by design. An error is more like there's an issue in the code that needs to be addressed like we're dividing a number the caller provided by zero.

By having the student create a custom exception that we check for, we can start to show that a more specific type makes handling exceptions easier.

exercises/practice/nucleotide-count/.meta/lib/example.dart
```diff
@@ -0,0 +1,17 @@
+class InvalidNucleotideException implements Exception {}
+
+class NucleotideCount {
+  Map<String, int> count(String strand) {
+    final result = {"A": 0, "C": 0, "G": 0, "T": 0};
+
+    for (final nucleotide in strand.split('')) {
+      if (!result.containsKey(nucleotide)) {
+        throw InvalidNucleotideException();
```
glennj, line 9: I'm not seeing where the custom exception is defined.

glennj: D'oh! How did I not see that?

glennj (approved):

PR #589, opened 2024-09-23
Add `strain`

BNAndras:


glennj:
@BNAndras The stub and the example are swapped.

BNAndras:
Oh jeez. This is why I shouldn’t port an exercise right before bed.

glennj (approved):

PR #595, opened 2024-10-04
Update SNIPPET.txt

hesham04Dev:
replace : by ; fixing the code err


glennj (approved):
Thanks for noticing this!

PR #600, opened 2024-10-15
remove WIP concept exercises from config.json: there is no WIP

glennj:
When concept exercises are actually in progress, this can be reinstated.


Stargator:
> When concept exercises are actually in progress

It was my understanding that Exercism was working through the tracks to develop the concept exercises.

These, I believe, were the initial attempt but the guidance for concept exercises kept shifting.

I believe working on these are more important than implementing more exercises from the problem-specifications repo.

glennj:
`configlet create` is useful for creating the exercise skeleton, so we don't need these anymore.

glennj:
@Stargator any more thoughts on this? I agree that writing a syllabus is very high value, but nobody is doing it.

Stargator:
@glennj, I have no issue with removing these. It doesn't look good on the website as they are incomplete.

Stargator (approved):

PR #625, opened 2025-06-26
Add `flower-field`, deprecating `minesweeper`

BNAndras:
https://forum.exercism.org/t/suggestion-deprecate-minesweeper-for-flower-field/17967/21

Could you label this as `rep/tiny`?


glennj:
While you're in here, can you add `https://stackoverflow.com/questions/tagged/dart` to the `.lycheeignore` file. I thought it was already there, but apparently not.

PR #650, opened 2026-03-19
Update exercise generator

BNAndras:
See https://forum.exercism.org/t/updates-to-the-exercise-generator/29635.

This creates a new test suite like below. The generator notably doesn't handle error property tests very well. I think doing a bulk update might not be appropriate. See https://github.com/exercism/dart/compare/main...BNAndras:bulk-update-example?expand=1 for an example of what all would be updated. There are a few spots where previous implementers diverged from the problem-specs so it might not be worth it. That kinda sucks because the main reason I wanted to update this generator was to use it to help regenerate test suites. However, even just fixing the parts related to new exercise creation would be useful since this has been broken for a while, and I worked around it.

```dart
import 'package:line_up/line_up.dart';

void main() {
  final lineUp = LineUp();

  group('LineUp', () {
    test('format smallest non-exceptional ordinal numeral 4', () {
      final result = lineUp.format('Gianna', 4);
      expect(result, equals('Gianna, you are the 4th customer we serve today. Thank you!'));
    }, skip: false);

...

    test('format exceptional ordinal numeral 123', () {
      final result = lineUp.format('Yma', 123);
      expect(result, equals('Yma, you are the 123rd customer we serve today. Thank you!'));
    }, skip: true);
  });
}
```

### DIFF
--- a/bin/create_exercise.dart
+++ b/bin/create_exercise.dart
@@ -3,7 +3,6 @@ import 'dart:io';
 
 import 'package:args/args.dart';
 import 'package:path/path.dart' show dirname;
-import 'package:yaml/yaml.dart';
 
 // Constants
 const _scriptFileName = 'create-exercise';
@@ -79,9 +78,9 @@ String testTemplate(String name) {
 import 'package:${snakeCase(packages[0])}/${snakeCase(packages[0])}.dart';
 import 'package:${snakeCase(packages[1])}/${snakeCase(packages[1])}.dart';
 
-final ${camelCase(name)} = ${pascalCase(name)}();
-
 void main() {
+  final ${camelCase(name)} = ${pascalCase(name)}();
+
   group('${pascalCase(name)}', () {
 $_testCasesString
   });
@@ -90,11 +89,10 @@ $_testCasesString
 }
 
 /// Generates the yaml code for a pubspec.yaml file.
-String pubTemplate(String name, String version) => '''
+String pubTemplate(String name) => '''
 name: '${snakeCase(name)}'
-version: $version
 environment:
-  sdk: '>=2.18.0 <3.0.0'
+  sdk: '>=3.2.0 <4.0.0'
 dev_dependencies:
   test: '<2.0.0'
 ''';
@@ -123,7 +121,7 @@ linter:
 
 /// Parses through the given test case (or group) in order to produce a String of code for the generated test suite.
 String testCaseTemplate(String exerciseName, Map<String, dynamic> testCase,
-    {bool firstTest = true, String returnType = ''}) {
+    {bool firstTest = true, String returnType = '', Set<String> excludedUuids = const {}}) {
   bool skipTests = firstTest;
 
   if (testCase['cases'] != null) {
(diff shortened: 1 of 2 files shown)


glennj:
I'll look at this when I have more time to concentrate. In the meantime there's some failing checks to look into.

glennj:
> I'll look at this when I have more time to concentrate. In the meantime there's some failing checks to look into.

```sh
dart pub get
dart run dart_style:format -i 0 -w -l 120 bin/  ## "fix" the formatting
dart analyze bin/ lib/ test/                    ## << reveals the problem
```

bin/create_exercise.dart
```diff
@@ -90,11 +89,10 @@ $_testCasesString
 }
 
 /// Generates the yaml code for a pubspec.yaml file.
-String pubTemplate(String name, String version) => '''
+String pubTemplate(String name) => '''
```
glennj, line 92: I see you removed a parameter here, but didn't in the create_exercise test

bin/create_exercise.dart
```diff
@@ -90,11 +89,10 @@ $_testCasesString
...
 name: '${snakeCase(name)}'
-version: $version
 environment:
-  sdk: '>=2.18.0 <3.0.0'
+  sdk: '>=3.2.0 <4.0.0'
```
glennj, line 95: You'll need to change the test's expected value too.

bin/create_exercise.dart
```diff
@@ -202,54 +206,74 @@ String _finalizeReturnType(String expected, String returnType) {
...
+      stderr.write('Warning: `configlet sync` exited with an error.\n');
+    }
+  }
+
+  // Generate tests after configlet sync so excluded UUIDs from tests.toml are up to date
+  _testCasesString = testCaseTemplate(exerciseName, specification, excludedUuids: _getExcludedUuids(exerciseDir));
+
+  Directory('${exerciseDir.path}/.meta/lib').createSync(recursive: true);
   Directory('${exerciseDir.path}/lib').createSync(recursive: true);
   Directory('${exerciseDir.path}/test').createSync(recursive: true);
```
glennj:
```suggestion
  Directory('${exerciseDir.path}/.meta/lib').createSync(recursive: true);
Directory('${exerciseDir.path}/lib').createSync();
Directory('${exerciseDir.path}/test').createSync();
```
The first directory creates the path, the other 2 don't need to be recursive.

glennj (approved):
I assume you tested it to create/update an exercise?

bin/create_exercise.dart
BNAndras: `test: '<2.0.0'`? That matches the version used in the exercises' pubspec.yaml files, but I might be missing something here.

BNAndras: Ooh, I didn't realize there was a test suite for this script. This comment makes sense now.

BNAndras:
```pwsh
PS C:\Users\anagy\Documents\GitHub\exercism-dart> dart analyze bin/ lib/ test/ ## << reveals the problem
Analyzing bin, lib, test...
No issues found!
```

